PR #141, opened 2014-09-03
Add filename_proc option to Factory generator

kreintjes:
Add a filename_proc option to the Factory generator for better control over the generated factory filename.

Main motivation for this pull request is the ability to use the singularized model_name in the factory filename. For the model Book I want to be able to call the factory book_factory.rb, while with the suffix option only books_factory.rb can be achieved. I believe the latter filename is not semantically correct, because the file defines the data for a single book.

With the filename_proc option people could of course also add prefixes or suffixes to the filename. Therefore this option could potentially replace the suffix option, although it is also possible to support them both for legacy reasons.

### DIFF
--- a/features/fixture_replacement_config.feature
+++ b/features/fixture_replacement_config.feature
@@ -115,3 +115,18 @@ Feature:
       | spec/factories/users_suffix.rb |
     Then the following files should not exist:
       | spec/factories/users.rb |
+
+  Scenario: Use a filename_proc with the Factory Girl generator
+    When I add "rspec-rails" as a dependency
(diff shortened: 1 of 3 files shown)


joshuaclayton:
@kreintjes can you rebase against master and push the update? I recently merged a fix to Travis in 8c16ae46e8a26d3415b16e8f086b675ac1d1284a

kreintjes:
@joshuaclayton Done!

joshuaclayton:
This has been merged in 5b17b5777b72028b2338c347e333a804d9adca9b

PR #144, opened 2014-10-18
Cleanup gemspec

igas:
* `test_files` don't used by rubygems anymore
* `executables` need only for contributors
* minimize gem size by including only needed files


joshuaclayton:
Resolved in a60c1c79f5039d0b2d2cbff484ba9260a16a8d04 and c789e9db8c9823d740a841c687645fe51c7135eb

PR #145, opened 2014-10-23
Change FactoryGirl::Generators::Base#explicit_class_option to output 1.9 Hashes

petedmarsh:
This changes the generator to use the 1.9-style Hash syntax for the `class` option. So, where previously the generator would output:

```ruby
FactoryGirl.define do
  factory :buttery_biscuit_base, :class => 'Buttery::BiscuitBase' do
    #...
  end
end
```

it will now output:

```ruby
FactoryGirl.define do
  factory :buttery_biscuit_base, class: 'Buttery::BiscuitBase' do
    #...
  end
end
```

I totally understand this is a very minor change and is ultimately down to personal preference - but it is something I would like :). FactoryGirl requires Ruby > 1.9 (https://github.com/thoughtbot/factory_girl/blob/master/factory_girl.gemspec#L17) so this shouldn't cause any issues in practice.

### DIFF
--- a/features/generators.feature
+++ b/features/generators.feature
@@ -15,7 +15,7 @@ Feature:
     Then the output should contain "test/factories/users.rb"
     And the output should contain "test/factories/namespaced_users.rb"
     And the file "test/factories/users.rb" should contain "factory :user do"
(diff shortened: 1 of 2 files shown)


joshuaclayton:
Thanks! Merged in 3a050736f3ea0b84e389caef5ac1bc8dc7b7b679

PR #156, opened 2015-02-11
Remove `setup` from gem's executables list

delphaber:
bin/setup script should be used in development. So `factory-girl-rails` should not declare `setup` as a gem executable file


delphaber:
sorry. duplicate

Issue #165, opened 2015-05-25
initializer clobbers configured definition_file_paths


paulwalker:
The initializer ignores any value already present in it's mapping to full Rails paths. This is deceptive as one cannot change the paths in configuration because of this. Instead, it should utilize whatever value is present and simply map those to full Rails paths.

PR #166, opened 2015-05-25
utilize configured definition_file_paths value in initializer closes …

paulwalker:
…#165

### DIFF
--- a/lib/factory_girl_rails/railtie.rb
+++ b/lib/factory_girl_rails/railtie.rb
@@ -10,11 +10,10 @@ module FactoryGirl
     end
 
     initializer "factory_girl.set_factory_paths" do
-      FactoryGirl.definition_file_paths = [
-        Rails.root.join('factories'),
-        Rails.root.join('test', 'factories'),
-        Rails.root.join('spec', 'factories')
-      ]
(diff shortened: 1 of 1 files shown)


lib/factory_girl_rails/railtie.rb
```diff
@@ -10,11 +10,10 @@ class Railtie < Rails::Railtie
     end
 
     initializer "factory_girl.set_factory_paths" do
-      FactoryGirl.definition_file_paths = [
-        Rails.root.join('factories'),
-        Rails.root.join('test', 'factories'),
-        Rails.root.join('spec', 'factories')
-      ]
+      paths = FactoryGirl.definition_file_paths || []
```
composerinteralia: I like the general idea here, but I don't know if this would totally solve the issue of allowing custom definition_file_paths. I think the initializer will run before any of your application code, so you wouldn't have a chance to customize these paths early enough to have an effect. We could maybe add a configuration option `config.factory_bot.definition_file_paths` to solve this.

I know it has been a while. Are you interested in working on this? I would be happy to help, or I can take over if you are no longer interested. If I don't hear from you I will look into this next week.

PR #179, opened 2015-11-14
Add spaces

knao124:
if we exec `rails generate model sample_model hoge:integer fuga:integer` then the generated factory file does not have correct indent spaces like following, 

```
FactoryGirl.define do
  factory :sample_model do
    hoge 1
fuga 1
  end

end
```

This PR will fix this.


djsegal:
+1

composerinteralia:
It looks like this got fixed with https://github.com/knao124/factory_girl_rails/commit/3488c767c76d6fe3d187c788ec158cf3d9649798#diff-37ad23751b9b72d7001b34c32168f983. Thanks!

PR #198, opened 2016-04-01
Relax version constraint to allow factory_girl 4.7

jaredbeck:
No idea if FGR supports FG 4.7 :D but let's see what CI says. #yolodev


joshuaclayton:
@jaredbeck I'm in the process of releasing 4.7.0 of FGR to address this actually - just working out some things on TravisCI currently. Thanks!

PR #207, opened 2016-04-29
Update Gemspec file to not use `git ls-files`

### LINKED ISSUE #177: Git dependency
I guess you already know about it, but I cannot see it anywhere in the documentation: this gem is dependent on Git (factory_girl_rails.gemspec uses ``git ls-files``). [Here](https://github.com/typus/typus/pull/422/files) is an example of how it could be solved.

ralinc:
The gem is dependent on Git (factory_girl_rails.gemspec uses git ls-files).
Some prod environments do not have git and if you try adding a gem through
git or path option in bundler, 'git ls-files' in the .gemspec fails.
Instead of using git, we could use pure Ruby code to list the gem files.

Fixes #177.

### DIFF
--- a/factory_girl_rails.gemspec
+++ b/factory_girl_rails.gemspec
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
     factory_girl and rails 3 (currently just automatic factory definition
     loading)}
 
-  s.files         = `git ls-files`.split("\n")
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@ralinc thanks - I've merged this as a44854dc0ca813e8b01912f6dfd868978e09858f

PR #208, opened 2016-05-08
Add $ to command line instruction

shinyaK14:
I've added $ mark into console instruction.
$ marks help us to understand which is console and which is Ruby's source code.
It's the same as [Rails README.md](https://github.com/rails/rails/blob/master/README.md)

**Before:**
```console
gem install factory_girl_rails
```
```ruby
gem 'factory_girl_rails'
```

**After:**
```console
$ gem install factory_girl_rails
```
```ruby
gem 'factory_girl_rails'
```

Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Github: http://github.com/thoughtbot/factory_girl_rails
 
 Gem:
 
-    gem install factory_girl_rails
+    $ gem install factory_girl_rails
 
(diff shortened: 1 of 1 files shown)


joshuaclayton:
Thank you @shinyaK14! I've merged this as 282b293d3fd23ef33f8fa4a27c9722c048d7377d.

PR #216, opened 2016-07-16
Updates ReadMe to recommend environments for gem

pareidoliax:
Recommends adding  `factory_girl_rails` to both your `:test` and `:development` groups so generators will work in development. As per issue  https://github.com/thoughtbot/factory_girl_rails/issues/88


joshuaclayton:
Thanks @pareidoliax! I've merged this as 4a7afc4cf508afa369b31a377f81b9700955dae9

PR #220, opened 2016-10-27
Update README.md

toomey8:
Changed links to https://thoughtbot.com/services/ruby-on-rails. Ready for review @thoughtbot/growth


composerinteralia:
Do we still want to make this change?

composerinteralia:
I just noticed we actually restructure the README a bit since this was open, and I don't this we need this change anymore.

PR #223, opened 2016-12-20
Replace Aruba's deprecated methods #before_cmd and #set_env.

otakumesi:
In `features/support/env.rb`.

There are Aruba's deprecated methods, So I replaced them to be recommended methods.

- `#before_cmd` to `#before(:command)`
- `#set_env` to  `#set_environment_variable`

Thanks.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -1,21 +1,26 @@
-require 'aruba/cucumber'
+require "aruba/cucumber"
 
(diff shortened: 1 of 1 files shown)


Otakumesi:
Sorry.
I modified the style guide violations.

I had do read CONTRIBUTING.md.

I will be careful.

PR #224, opened 2016-12-21
Update Ruby versions on Travis CI

tricknotes:


composerinteralia:
It looks like this got addressed in #272. Thanks for the PR!

Issue #225, opened 2017-01-05
log in gem


znz:
Some log files exist in released gem.

```
% (cd $(gem env gemdir)/gems && ls factory_girl_rails-4.8.0/**/*.log)
factory_girl_rails-4.8.0/tmp/aruba/testapp/log/development.log
factory_girl_rails-4.8.0/tmp/aruba/testapp/log/test.log
```

PR #240, opened 2017-10-18
Don't include specs and features in a gem

meganemura:
And include only version controlled files by git.

This change prevents #225.

### DIFF
--- a/factory_girl_rails.gemspec
+++ b/factory_girl_rails.gemspec
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
     factory_girl and rails 3 or newer (currently just automatic factory definition
     loading)}
 
-  s.files         = Dir['**/*'].keep_if { |file| File.file?(file) }
+  s.files         = `git ls-files`.split("\n").reject { |f| f.match(%r{^(spec|features)/}) }
   s.require_paths = ["lib"]
   s.executables   = []
   s.license       = "MIT"


factory_girl_rails.gemspec
```diff
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
     factory_girl and rails 3 or newer (currently just automatic factory definition
     loading)}
 
-  s.files         = Dir['**/*'].keep_if { |file| File.file?(file) }
+  s.files         = `git ls-files`.split("\n").reject { |f| f.match(%r{^(spec|features)/}) }
```
mjankowski: FYI, bundler/rubygems will raise a warning (or maybe not work at all? I forget which version this started in) with the usage of running subcommands like this.

I think that ignoring spec and features for gem packaging makes sense, but can you update this to remove the `git ls-files...` addition?

mjankowski: Yeah, agreed that it's popular ... I could have sworn that I saw something advising against using it. Maybe I'm thinking of something else?!

meganemura: I found the related issue at bundler/bundler#5810. This issue describes installing a gem from source in an environment without git is impossible. bundler core-developer replied in following comment.
https://github.com/bundler/bundler/issues/5810#issuecomment-310480516

composerinteralia:
I would like to get this in before our next release, but I don't think we have quite the correct solution. We are still including all the Appraisal Gemfiles, and some other things we probably don't want. What about something like:

```
`git ls-files lib LICENSE README.md`.split($RS)
```

Am I missing any other files that should be included?

composerinteralia:
Thanks for your help on this. I made a few changes to your commit and opened #286.

PR #244, opened 2017-10-21
Add factory_bot_rails transition instructions

Euraldius:


README.md
```diff
@@ -6,6 +6,18 @@ instances, attribute hashes, and stubbed objects), and support for multiple
...
+
+Transitioning is a two step process:
+
+1) Replace `factory_girl_rails` with `factory_bot_rails` in your Gemfile
+    and `bundle install`
+
+2) Rename all `FactoryGirl` to `FactoryBot` and `factory_girl` to `factory_bot`
+    in your codebase. This command may be helpful:
+
+    `ag -0l FactoryGirl spec | xargs -0 sed -i 's/FactoryGirl/FactoryBot/g'`
```
joshuaclayton: I wonder if we should specify that this only happen in `spec` - specifically, I'm thinking about situations like `rake dev:prime` where FG is used that would be missed.

joshuaclayton: Thoughts on backing this out to the top-level indentation and using triple-backticks?

```
```sh
```

README.md
```diff
@@ -6,6 +6,18 @@ instances, attribute hashes, and stubbed objects), and support for multiple
 factories for the same class (`user`, `admin_user`, and so on), including factory
 inheritance.
 
+### Transitioning from factory\_girl\_rails?
+
+Transitioning is a two step process:
+
+1) Replace `factory_girl_rails` with `factory_bot_rails` in your Gemfile
```
joshuaclayton: Possible to use Markdown for these bullet-points?

Euraldius: I think this is Markdown? Github's preview for this file seems to be parsing it as valid Markdown

joshuaclayton: TIL! That's pretty cool, I think this is easier to read.

README.md
```diff
@@ -6,6 +6,21 @@ instances, attribute hashes, and stubbed objects), and support for multiple
...
+    and `bundle install`
+
+2) Rename all `FactoryGirl` to `FactoryBot` and `factory_girl` to `factory_bot`
+    in your codebase.
+
+This command will perform a `FactoryGirl` to `FactoryBot` rename within your
+`spec/` directory:
+
+```ag -0l FactoryGirl spec | xargs -0 sed -i 's/FactoryGirl/FactoryBot/g'```
```
ekohl: Not every user might have ```ag``` installed. ```grep``` is more portable.

mjankowski:
We wound up referencing this file in the blog post - https://github.com/thoughtbot/factory_bot/blob/4-9-0-stable/UPGRADE_FROM_FACTORY_GIRL.md

Should we...

- Keep cleaning up this PR and merge something into the README as well?
- Close this PR in favor of a single UPGRADING.md file, maybe use that one in `-stable` as base?
- Put some combination of that file from `-stable` and this PR into the README?

Not sure.

joshuaclayton:
@mjankowski I think having `master` reference the upgrade doc you'd linked to, and have it capture both fg/fgr -> fb/fbr upgrades, would work well (since there are no specific steps for fgr -> fbr).

PR #253, opened 2017-12-31
CI against Ruby 2.5.0

ttanimichi:


composerinteralia:
It look like we ended up adding 2.5 in #272. Thanks for your PR!

PR #259, opened 2018-02-28
Switch to vector badges in README

rmm5t:
because raster makes eyes bleed on retina displays :sunglasses:


composerinteralia:
This is a duplicate of #221, which I just merged as 6604cc6a5229dff8184ec2863123c58393b4dfe5. Thanks!

PR #264, opened 2018-03-09
Run the factory definition search after ActiveRecord is loaded

vovimayhem:
This PR makes the Railtie to wait for ActiveRecord to be loaded before running the `FactoryBot.find_definitions` routine, which currently may load models before ActiveRecord is loaded, causing problems on some cases (i.e. doorkeeper-gem/doorkeeper#1043), specially when referencing ActiveRecord models in the `class` option inside a factory definition.

### DIFF
--- a/lib/factory_bot_rails/railtie.rb
+++ b/lib/factory_bot_rails/railtie.rb
@@ -18,7 +18,7 @@ module FactoryBot
     end
 
     config.after_initialize do
-      FactoryBot.find_definitions
+      ActiveSupport.on_load(:active_record) { FactoryBot.find_definitions }
 
       if defined?(Spring)
         Spring.after_fork { FactoryBot.reload }
--- a/lib/factory_bot_rails/railtie.rb
+++ b/lib/factory_bot_rails/railtie.rb
@@ -4,6 +4,13 @@ require 'rails'
 
 module FactoryBot
   class Railtie < Rails::Railtie
+    ORMS_USING_LOAD_HOOKS = {
+      'Mongoid' => :mongoid,
+      'SequelRails' => :sequel,
+      'MongoMapper' => :mongo_mapper,
+      'ActiveRecord' => :active_record,
+      # datamapper doesn't seem to use load hooks...
+    }.freeze
 
     initializer "factory_bot.set_fixture_replacement" do
       FactoryBotRails::Generator.new(config).run
@@ -18,10 +25,21 @@ module FactoryBot
     end
 
     config.after_initialize do
-      ActiveSupport.on_load(:active_record) { FactoryBot.find_definitions }
+      find_definitions_after_orm_load
 
       if defined?(Spring)
         Spring.after_fork { FactoryBot.reload }
+      end
+    end
+
+    def self.find_definitions_after_orm_load
+      return FactoryBot.find_definitions unless hook_name = orm_load_hook_name
+      ActiveSupport.on_load(hook_name) { FactoryBot.find_definitions }
+    end
+
+    def self.orm_load_hook_name
+      ORMS_USING_LOAD_HOOKS.each do |orm_module_name, hook_name|
+        return hook_name if orm_module_name.safe_constantize
       end
     end
   end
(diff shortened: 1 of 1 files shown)


vovimayhem:
@Euraldius

nbulaj:
Hi @vovimayhem . I think `factory_bot` can be used not only with ActiveRecord, but also with Mongoid, mongomapper, Sequel or any other ORM. Most of them are have ActiveSupport loading hooks (for Sequel - only by using `sequel-rails` gem). By introdusing this change you can break the specs for the projects using mentioned gems.

Btw, I think it is a good change request. All you need to do is to refactor it with support for any popular ORM with fsllback to simple call of `FactoryBot.find_defenitions` in case no ORM found.

vovimayhem:
Got it @nbulaj ! Let me see what can I do.

vovimayhem:
Spent a time looking at https://gist.github.com/booch/8925423, went through each project's repos (all of datamapper's repos!) searching for `run_load_hooks`... The only ones I found using load hooks are the ones @nbulaj mentioned. Anyways, if someone else gets stuck with the same problem, it will be quite simple to add the module + hook entry on the dictionary I added to the railtie.

vovimayhem:
Are jruby tests failing for everybody else? Or just for me???

composerinteralia:
Thanks for the PR! I don't quite understand why we need this. I have referred to AR models just fine in other projects without running into this. Is this an incompatibility between factory_bot and doorkeeper, or can we reproduce the behavior without the doorkeeper gem?

vovimayhem:
@composerinteralia Man! I don't know... it's been a while! Let me check what was I doing and why I found it...

vovimayhem:
OK as I remember (after reading all the stuff), when the `factory_bot_rails` railtie is initialized, it runs the "find factory definitions" *before* `ActiveRecord` gets initialized, missing models that are added to the project *after* `ActiveRecord` gets initialized.

I've come across a couple (or more) gems that wait for `ActiveRecord` to initialize before adding the models to the app - such is the case of `doorkeeper`. Given it is a fairly popular one, we should consider adding this compatibility.

We can reproduce this by requiring model files inside a "lazy_load" hook, which is the way doorkeeper models are loaded into the app (See https://github.com/doorkeeper-gem/doorkeeper/blob/85ae24091816b96d398192f19138304a39ff77c3/lib/doorkeeper/orm/active_record.rb#L32) and watching how the `factory_bot_rails` misses those models (as it's the case for `Doorkeeper::AccessToken`)

vovimayhem:
@composerinteralia should I include a test for this behavior?

etgrieco:
@vovimayhem I tried implementing this fix as a monkey-patch short-term, but it doesn't resolve my issue. The error I get is:
```
Circular dependency detected while autoloading constant User
```
(where User is the first model constant being loaded in specs).

On the other hand, not including `railties.rb` is starting to create other errors now where `FactoryBot` is an uninitialized constant as well. For now, I monkey-patched on the Doorkeeper end to revert the lazy-loading feature (see: https://github.com/doorkeeper-gem/doorkeeper/issues/913) temporarily, since it's becoming paramount to implement the `4.4.0` security update and I can't wait for a solution to be found on the FactoryBot end.

For other users running into this issue, this means copy-pasting `lib/doorkeeper/orm/active_record.rb` with the contents of https://github.com/doorkeeper-gem/doorkeeper/blob/f685963516882b6c2bfa8ff9bf6316d78e2f3957/lib/doorkeeper/orm/active_record.rb

vovimayhem:
@etgrieco Can U give us details about how did you monkey-patched the railtie?

etgrieco:
@vovimayhem I actually just tried to reimplement your solution (copy-pasting your code to `lib/factory_bot_rails/railties` and reimplemented it using string references to classes versus the class constants themselves and that fixed the issue universally. No need to monkey-patch doorkeeper's active_record module any longer! I.e. `factory :admin, class: 'User' do ...` vs.  `factory :admin, class: User do ...`

composerinteralia:
If using a string works alright, I would prefer to close this PR and open a documentation issue on factory_bot.

I appreciate the work you have done here, but I fear we will end up in a race to be the last to load. Is there anything to guarantee our AR on_load hook would run after doorkeeper's, for example, or would we have to make sure factory_bot appears later in the Gemfile? I think passing a string for the class is cleaner and more reliable.

vovimayhem:
@composerinteralia Agree.

composerinteralia:
Opened https://github.com/thoughtbot/factory_bot/issues/1188

PR #275, opened 2018-06-12
Update README.md

cpapidas:


composerinteralia:
Thanks! Right now it looks like this only applies if you are upgrading from the old gem name:

screen shot 2018-06-12 at 6 35 47 am

Could we move up into the paragraph about `factory_bot`?

Also, I think https://github.com/thoughtbot/factory_bot#documentation is a better link than what I gave you before, since it also includes a link to docs and to an introductory video.

cpapidas:
yea I ll fix it, thanks

PR #276, opened 2018-06-15
Test against Rails 5.1 and 5.2

composerinteralia:
- Add Rails 5.1 and 5.2 to Appraisal
- Avoid running combinations we know are going to fail by moving to
"exclude" instead of "allow_failures"


composerinteralia:
Didn't run travis when I made this PR via hub :-(

PR #279, opened 2018-06-22
Add version and SemVer stability badges to README

greysteil:
First of all, thanks for factory_bot! I use it everywhere and my tests are way cleaner for it.

Would you be up for adding a badge showing how stable / bug-free new releases are? I realised we've got the data to show that at Dependabot, so I threw together the below:

[Gem Version](http://badge.fury.io/rb/factory_bot_rails) [SemVer](https://dependabot.com/compatibility-score.html?dependency-name=factory_bot_rails&package-manager=bundler&version-scheme=semver)

The gem version badge is totally standard. The semver stability one is based on all of the updates Dependabot has done for factory_bot users - if you click through there's a description of how it's calculated, but basically it's the percentage of SemVer compatible updates for factory_bot that users of the gem have done which have passed CI.

The idea is that together they give exactly the right information for users to decide what version requirement to put in their Gemfile (as suggested in your blog post [here](https://robots.thoughtbot.com/a-healthy-bundle)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# factory_bot_rails [![Build Status][ci-image]][ci] [![Code Climate][grade-image]][grade]
+# factory_bot_rails [![Build Status][ci-image]][ci] [![Code Climate][grade-image]][grade] [![Gem Version][version-image]][version]
 
 [factory_bot][fb] is a fixtures replacement with a straightforward definition
 syntax, support for multiple build strategies (saved instances, unsaved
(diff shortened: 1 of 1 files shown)


composerinteralia:
I went back and forth on this, but I think I am going to pass on the semver stability badge. I have heard good things about Dependabot, but I have concerns about whether that percentage accurately represents what it says it does (CI can fail for all sorts of other reasons), and about how useful it is to people using the library (I would normally just update and trust my test suite to let me know if anything is broken). The gem version badge seems fine to me; if you want to update the PR to include just that I will merge.

greysteil:
👍 - updated to just be the version number badge, so this should be easy to merge.

I also put together a different version of the badge, that just pulls in updates to the latest version, that the folks at Webpack are using. It looks like this and the idea is that it highlights if the latest release breaks SemVer, just in case you think that would be more helpful:

[](https://dependabot.com/compatibility-score.html?dependency-name=factory_bot_rails&package-manager=bundler&version-scheme=semver&new-version=latest)

Finally, I'm going to expand the show screen for that badge to include links to any test failures (for public repos) to make it easy for maintainers to debug any issues if there *are* and accidental breaking  changes / bugs in a release.

composerinteralia:
Thanks!

Issue #192, opened 2016-02-17
find_definitions is loading test code in development


betesh:
Some of our factories (in `spec/factories/`) contain code that modifies our model definitions.

There are a few reasons we do this:
- Some functionality is only needed in factories (setting a private attribute directly without resorting to metaprogramming)
- Some DB tables don't have any ActiveRecord model class because we only interface with them via java (We are using jRuby)

Such code is not intended to be loaded in development mode and when it is loaded, it actually interferes with definitions (or absence of definitions) in `app/models`.

However, it **is** being loaded by [FactoryGirl::Railtie](https://github.com/thoughtbot/factory_girl_rails/blob/v4.6.0/lib/factory_girl_rails/railtie.rb#L21).

We are using this monkey-patch (with rails 3.2, factory_girl_rails 4.3.0) to make factory_girl_rails only be loaded when running a generator:

```
# Gemfile
 group :development do
   gem "factory_girl_rails", require: false
 end


# config/initializers/factory_girl_rails.rb
class ::Rails::Engine
  def load_generators_with_factory_girl_rails(app=self)
    require "factory_girl_rails"
    # Note that it's a little late to be loading the gem,
    # as initializers have already run.
    # So we need to run the initializer ourselves.
    FactoryGirl::Railtie.instance.run_initializers(:default, Rails.application)
    load_generators_without_factory_girl_rails(app)
  end

  alias_method_chain :load_generators, :factory_girl_rails
end
```

However, the ideal solution would be to make the after_initialize block in the railtie conditional on whether a generator is running.

composerinteralia:
Closing this issue, since we added a new configuration option that will allow you to avoid loading factory definitions for cases like this.

PR #281, opened 2018-07-14
Allow reloading of factory definitions

### LINKED ISSUE #236: Calling `reload!` in rails console should reload factory definitions
When I make changes to my rails code, I can load the new code into my active console session by calling `reload!` in the console. This does not reload factory girl definitions, however. That must be done manually.

It'd be nice if this happened automatically. I think (but am not sure) that it could be done with a `to_prepare` callback in an initializer inside the railtie. I also think that it *may* supercede the special handling we give spring in that railtie, but I'm not positive.

composerinteralia:
Closes #236

This commit uses ActiveSupport's FileUpdateChecker to allow reloading FactoryBot of definitions whenever a file in `FactoryBot.definition_file_paths` gets updated. This is similar to reloading for [I18n](https://github.com/rails/rails/blob/ced104d57997c7bceef3d1e6c8a713431363c3bb/activesupport/lib/active_support/i18n_railtie.rb#L60-L70) and [react rails](https://github.com/reactjs/react-rails/blob/83b6175460b6fd19d667854ebea4777d8c73705a/lib/react/rails/railtie.rb#L35-L41).

This allows us to get rid of any Spring-specific logic in the railtie, since [Spring hooks into the application reloader](https://github.com/rails/spring/blob/0c711ff10b4ad7dcc34282b5e08573c2ce1e668a/lib/spring/application.rb#L161).

This partly solves #211, since we no longer call `FactoryBot.reload` at all in `after_initialize`. Instead, we will only call it when one of the files in `definition_file_paths` gets updated. I say it partly solves #211 because once a definition file gets updated, reloading would still give warnings about redefining any constants in the definition files. I wonder how common it is to define constants in the same file as factory definitions. It's probably better to keep constants in the autoload path, allowing Rails to handle unloading and reloading them in development. I would want to see some specific examples before worrying too much about it.

I would also like to offer a better way to configure the definition file paths (see #165 and #166) and possibly an option to opt out of loading definitions at all (could help with issues like #192 and maybe #211 as well).

A couple of quirks here:
* the to_prepare block could potentially get [called multiple times](https://github.com/rails/rails/issues/28108). It shouldn't matter, since `execute_if_updated` is a no-op if no updates have been made.

* I noticed that the first time I call `reload!` in the console the factory definitions get reloaded even when I made no updates to the definition files. I think it is related to [this](https://github.com/rails/rails/blob/f7c5a8ce262cf16ad8d7c0b0d631a4b88afec414/activesupport/lib/active_support/evented_file_update_checker.rb#L18). After the first call `reload!` works as expected, only reloading the factory definitions if the definition files were updated.

* Rails uses execute rather than execute_if_updated for the [route
reloader](https://github.com/rails/rails/blob/master/railties/lib/rails/application/finisher.rb#L133) and for the [watchable file reloader](https://github.com/rails/rails/blob/master/railties/lib/rails/application/finisher.rb#L173). This means that changes to factory definitions will cause reloading of
the whole application. I think this is fine.


composerinteralia:
Hm, these tests are turning out to be flaky on Travis. They worked locally with the same combinations of Ruby and Rails. 😢 I guess I have my next project! Still feel free to look over and let me know whether you think the general approach makes sense.

composerinteralia:
@derekprior any interest in reviewing this?

Issue #180, opened 2015-11-18
Difficult to access engine factories from main application


crystalneth:
I spent a lot of time tracking down how to do this, and the method I found seems obscure and wrong.

I want to use an engine's factories within the main application, e.g. FactoryGirl.create(:engine_product).

Adding to the engine gem's files and test_files does not help. I could not figure out how to to require the factory files.

Ultimately, I ended up adding what's below to the engine file, something I tracked down buried in a merge comment. Should there be a more supported way of exposing engine factories?

``` ruby
    initializer "engine.factories", :after => "factory_girl.set_factory_paths" do
      FactoryGirl.definition_file_paths << File.expand_path('../../../spec/factories', __FILE__) if defined?(FactoryGirl)
    end
```

composerinteralia:
https://github.com/thoughtbot/factory_bot_rails/pull/149 adds this to the README

composerinteralia:
Closing this issue, since we added a new configuration option (documented in the README) for fbr 5 that should make this easier.

PR #289, opened 2018-09-10
 Introduce definition file path configuration

### LINKED ISSUE #192: find_definitions is loading test code in development

composerinteralia:
Currently the only way to customize the factory definition file
paths is to do it in an initializer right after the
"factory_bot.set_factory_paths" initializer. With this PR,
we can customize factory definition file paths by
setting `config.factory_bot.definition_file_paths`
in config/application.rb or the appropriate environment file.

We should be able to close #149 and #180 as well, since using this
configuration can replace the initializer solution for sharing factories
in an engine (the initializer solution will still work, but with the new
configuration you don't need to know anything about the
fbr railtie).

This will also allow us to close #192, since we can use this
configuration with an empty array to disable automatic loading
of factories in development.

I added documentation for the definition_file_paths configuration,
and also restructured and expanded the rest of the README documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,7 @@ Check out the [guide](https://github.com/thoughtbot/factory_bot/blob/4-9-0-stabl
 
 ## Rails
 
-factory_bot_rails provides Rails integration for [factory_bot][fb].
-
-Currently, automatic factory definition loading is the only Rails-specific feature.
+factory\_bot\_rails provides Rails integration for [factory_bot][fb].
 
 Supported Rails versions are listed in [`Appraisals`](Appraisals). Supported
 Ruby versions are listed in [`.travis.yml`](.travis.yml).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -37,39 +35,101 @@ group :development, :test do
 end
 ```
 
-Generators for factories will automatically substitute fixture (and maybe any other
-`fixture_replacement` you set). If you want to disable this feature, add the
-following to your application.rb file:
+You may want to configure your test suite to include factory\_bot methods; see
+[configuration](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#configure-your-test-suite).
```
sidraval: Is there is a way to make this a relative link, so e.g. if someone is on a different branch/tag, they don't get routed to the master documentation?

README.md
```diff
@@ -37,39 +35,101 @@ group :development, :test do
...
+spec/factories.rb
+factories/*.rb
+test/factories/*.rb
+spec/factories/*.rb
+```
+
+You can configure this with:
+
+```ruby
+config.factory_bot.definition_file_paths = ["custom/factories"]
```
sidraval: Does it make sense to indicate what file this should be placed in, or is it clear/mentioned elsewhere?

README.md
composerinteralia: It needs to work on RubyDocs as well, and I haven't found a good way to make relative links work there.

PR #290, opened 2018-09-12
Remove unneeded conditional for app_generators

composerinteralia:
Configuration#generators was deprecated in favor of
Configuration#app_generators back in Rails 3.1 (https://github.com/rails/rails/commit/04cbabb0a0206553d7b474ee7a191d19957048fe).
Since we started work on factory_bot_rails 5, which only supports Rails
4.2+, we no longer need the conditional.

I also updated the development Gemfile, since it was out of date,
and replaced a nested if/else with elsif.

### DIFF
--- a/lib/factory_bot_rails/generator.rb
+++ b/lib/factory_bot_rails/generator.rb
@@ -5,11 +5,7 @@ require 'factory_bot_rails/generators/null_generator'
 module FactoryBotRails
   class Generator
     def initialize(config)
-      @generators = if config.respond_to?(:app_generators)
-                      config.app_generators
-                    else
-                      config.generators
-                    end
+      @generators = config.app_generators
     end
 
     def run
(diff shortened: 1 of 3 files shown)


lib/factory_bot_rails/generator.rb
```diff
@@ -19,12 +15,10 @@ def run
...
         Generators::NullGenerator
+      elsif test_framework == :rspec
+        Generators::RSpecGenerator
       else
-        if test_framework == :rspec
-          Generators::RSpecGenerator
-        else
-          Generators::NonRSpecGenerator
-        end
+        Generators::NonRSpecGenerator
```
sidraval: tbh I kind of like the previously `if/else` because the `elsif` is on first glance unrelated to to factory bot being disabled... does that make sense?

Anyway this is just preference, but figured I'd bring it up in case 💅

composerinteralia: I changed because RuboCop was yelling about nested conditionals. I will change it back, since it is not related to this PR at all.

PR #291, opened 2018-09-18
Add RuboCop configuration

composerinteralia:
This is similar to the work Christian Bruckmayer is doing over on
factory_bot. I read that Hound doesn't support the
`inherited_from` key, so I put all of the todos
from the auto-generated `.rubocop_todo.yml` straight into the
`.rubocop.yml`.


Gemfile
```diff
@@ -18,3 +18,5 @@ gem 'therubyrhino', platforms: :jruby
 gem 'jruby-openssl', platforms: :jruby
 
 gem 'sqlite3', platforms: :ruby
+
+gem "rubocop", require: false
```
ChrisBr: Please note that according to this documentation, HoundCI is using version 0.54 of rubocop:
http://help.houndci.com/configuration/rubocop

Maybe it is the reason that HoundCI is still pending (so it crashed because of an incompatible configuration?)

composerinteralia:
Thanks for the review @ChrisBr. Locked version to 0.54.0, and inherited from the same organization-wide `.rubocop.yml` we are using on Hound, so this should finally run the same locally as it does on Hound!

Issue #293, opened 2018-09-19
Gradually fix RuboCop todos


composerinteralia:
We added a .rubocop.yml in https://github.com/thoughtbot/factory_bot_rails/commit/29249b375168e06d9d8f23ea365f511cb63a00c6 with a bunch of todos. It would be nice to gradually fix them. I would like to do it one at a time so nobody has to review a giant PR. 😸

petestreet:
Hi @composerinteralia! I made a few PRs, grouped according to specific files or directories that were flagged by rubocop: https://github.com/thoughtbot/factory_bot_rails/pull/295, https://github.com/thoughtbot/factory_bot_rails/pull/296, https://github.com/thoughtbot/factory_bot_rails/pull/297, https://github.com/thoughtbot/factory_bot_rails/pull/298, and https://github.com/thoughtbot/factory_bot_rails/pull/299

A few notes: 
I wanted to get feedback before addressing `Layout/IndentHeredoc`. Rubocop suggests adding an additional dependencies such as ActiveSupport [link] for sub-2.3 Ruby versions, and using the `<<~` syntax otherwise, and I wasn’t sure which approach made more sense in the context of this gem. 

I also haven’t addressed `Metrics/LineLength` at all yet, but would be happy to do that after the other PRs are in. 

There are also a few PR-specific caveats that I made mention of for https://github.com/thoughtbot/factory_bot_rails/pull/296 and https://github.com/thoughtbot/factory_bot_rails/pull/299. 

After the PRs are in, I can go back and clear our the empty cops from `.rubocop.yml` 

Thanks!

PR #239, opened 2017-08-07
test with ruby-head on travis

naofumi-fujii:

PR #297, opened 2018-09-25
Address todos generated by rubocop for files in the features/ directory.

petestreet:
This partially addresses [#239](https://github.com/thoughtbot/factory_bot_rails/issues/293) by deleting references to files in the `features/` directory from `.rubocop.yml` and fixing the subsequent warnings.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -1,21 +1,23 @@
-require 'aruba/cucumber'
+require "aruba/cucumber"
 
-PROJECT_ROOT = File.expand_path(File.join(File.dirname(__FILE__), '..', '..')).freeze
+PROJECT_ROOT = File.expand_path(File.join(File.dirname(__FILE__), "..", "..")).freeze
 
 Aruba.configure do |config|
   config.exit_timeout = Integer ENV.fetch("ARUBA_TIMEOUT") { 120 }
 end
 
-Aruba.configure do |config|
-  config.before_cmd do |cmd|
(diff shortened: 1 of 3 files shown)


features/support/env.rb
```diff
@@ -1,21 +1,23 @@
...
+PROJECT_ROOT = File.expand_path(File.join(File.dirname(__FILE__), "..", "..")).freeze
 
 Aruba.configure do |config|
   config.exit_timeout = Integer ENV.fetch("ARUBA_TIMEOUT") { 120 }
 end
 
-Aruba.configure do |config|
-  config.before_cmd do |cmd|
-    set_env('JRUBY_OPTS', "-X-C #{ENV['JRUBY_OPTS']}") # disable JIT since these processes are so short lived
+if RUBY_PLATFORM == "java"
```
composerinteralia: We actually stopped testing against jruby in https://github.com/thoughtbot/factory_bot/commit/9f1b32e07100d9a7da5439532b14cb7c349a2563#diff-354f30a63fb0907d4ad57269548329e3, so I think we can remove all of this.

petestreet: @composerinteralia I can remove this entire block, but would that cause any issues given the existing support for jruby-specific gems? https://github.com/thoughtbot/factory_bot_rails/blob/master/Gemfile#L15-L18

composerinteralia: These are all test-only files that aren't included in the gem (see https://github.com/thoughtbot/factory_bot_rails/blob/master/factory_bot_rails.gemspec#L13), so removing them wouldn't have any effect on people using the gem. 

But I guess it may make it harder for someone using jruby to contribute, so we can leave it.

PR #306, opened 2018-11-30
Add newline between factories in same file

### LINKED ISSUE #305: Missing newline when generating factories in a single file
`rails generate model robot name:string`

In an existing `factories.rb`:

```rb
FactoryBot.define do
  factory :robot do
    name { "MyString" }
  end
  factory :human do
    #...
  end
end
```

generates the factory at the top of the file without any space between the `robot` factory and an existing `human` factory. It would be better if there were a newline there:

```rb
FactoryBot.define do
  factory :robot do
    name { "MyString" }
  end

  factory :human do
    #...
  end
end
```

composerinteralia:
Fixes #305 

When generating factories in the same file (for example
spec/factories.rb), before this PR we didn't put newlines between the
factories. So if you generated a user factory, then a robot factory, you
would end up with something like this:

```rb
FactoryBot.define do
  factory :robot do
    name { "MyString" }
  end
  factory :user do
    name { "MyString" }
  end
end
```

With this PR, generating those factories will now look like:

```rb
FactoryBot.define do
  factory :robot do
    name { "MyString" }
  end

  factory :user do
    name { "MyString" }
  end

end
```

Note that the final newline is not ideal, but it will only be added when
generating the first factory, whereas the missing newline after
factories was happening every time we generated a new factory. So
with this PR we will only need to fix a whitespace style error once, rather
than N-1 times. We can could open another issue to fix the extra newline
at the end, but I think it may be more complicated than it sounds.

Co-authored-by: Nick Sanford <me@nicksanford.io>

### DIFF
--- a/features/generators.feature
+++ b/features/generators.feature
@@ -34,12 +34,18 @@ Feature:
       end
       """
     And I run `bundle exec rails generate model User name:string` with a clean environment
+    And I run `bundle exec rails generate model Robot name:string` with a clean environment
     Then the file "test/factories.rb" should contain exactly:
       """
       FactoryBot.define do
+        factory :robot do
+          name { "MyString" }
+        end
+
         factory :user do
           name { "MyString" }
         end
+
       end
       """
 
@@ -51,4 +57,4 @@ Feature:
       end
       """
     And I run `bundle exec rails generate model User name:string` with a clean environment
-    Then the file "test/factories.rb" should contain "factory :user do"
+    Then the file "test/factories/users.rb" should not contain "factory :user do"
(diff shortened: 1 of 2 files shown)


features/generators.feature
```diff
@@ -51,4 +57,4 @@ Feature:
       end
       """
     And I run `bundle exec rails generate model User name:string` with a clean environment
-    Then the file "test/factories.rb" should contain "factory :user do"
+    Then the file "test/factories/users.rb" should not contain "factory :user do"
```
composerinteralia: Not strictly related to this PR, but this expectation didn't match Scenario description.

PR #307, opened 2018-12-01
Allow using custom template for the generator

### LINKED ISSUE #147: Quitting template usage is wrong
1. Standard rails generator behavior is asking the action in the case the file exists. So the generator should not check for the existence.
2. Quitting template usage breaks the use of more sophisticated templates like mine `factory_girl_fixtures_template`.
3. Quitting template usage breaks the standard rails generators workflow.
4. The factory generated with the new code is buggy breaking the ruby style.

So, please, restore back the templates directory in the gem and use of it so that the people that use custom templates for the FG, can continue explore them in the latest versions.

### LINKED ISSUE #148: rails generate factory_girl:model no longer uses factory_girl/model/fixtures.erb file?
I have the impression that pull request "Do not generate new file if `factories.rb` exists #135" also removed the possibility to have a custom fixtures.erb file in a project's lib/templates/factory_girl/model folder, as it used to be:

file lib/generators/factory_girl/model/model_generator.rb

```
      def create_fixture_file
  -        filename = [table_name, filename_suffix].compact.join('_')
  -        template 'fixtures.erb', File.join(options[:dir], "#{filename}.rb")
  +        if File.exist?(factories_file)
  +          insert_factory_into_existing_file
  +        else
  +          create_factory_file
  +        end
       end
```

composerinteralia:
PR #135 introduced the ability to generate factories in one file, but it
also removed the ability to use custom templates for generating
factories. This PR adds back the ability to use custom templates when
generating factories in separate files. I doubt this was a heavily used
feature, but it has come up a few times and it I think supporting it
simplifies the generator.

Closes #147
Closes #148
Closes #193

PR #193 basically reverted #135, which is undesirable. thoughtbot
projects generally prefer putting all the factories in one file.

Co-authored-by: Nick Sanford <me@nicksanford.io>

### DIFF
--- a/features/generators.feature
+++ b/features/generators.feature
@@ -7,10 +7,10 @@ Feature:
     Given I successfully run `bundle exec rails new testapp`
     And I cd to "testapp"
     And I add "factory_bot_rails" from this project as a dependency
+    And I run `bundle install` with a clean environment
 
   Scenario: The factory_bot_rails generators create a factory file for each model if there is not a factories.rb file
-    When I run `bundle install` with a clean environment
-    And I run `bundle exec rails generate model User name:string age:integer` with a clean environment
+    When I run `bundle exec rails generate model User name:string age:integer` with a clean environment
     And I run `bundle exec rails generate model Namespaced::User name:string` with a clean environment
     Then the output should contain "test/factories/users.rb"
     And the output should contain "test/factories/namespaced_users.rb"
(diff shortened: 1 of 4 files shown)


features/generators.feature
```diff
@@ -44,11 +42,23 @@ Feature:
...
+    When I write to "lib/templates/factory_bot/model/factories.erb" with:
+      """
+      <%= "Custom factory definition" %>
+      """
+    And I run `bundle exec rails generate model User` with a clean environment
+    Then the file "test/factories/users.rb" should contain exactly:
+      """
+      Custom factory definition
+      """
+
```
tabfugnic: Multiple lines at the end of this file.

lib/generators/factory_bot/model/templates/factories.erb
```diff
@@ -0,0 +1,3 @@
+FactoryBot.define do
+<%= factory_definition.chomp %>
```
tabfugnic: should this be indented?

in a followup(previous?) PR you use `rstrip` instead of `chomp` should this be using `rstrip`?

composerinteralia: It shouldn't be indented because `factory_definition` takes care of the indentation. But I agree it looks a bit strange, so I will open a separate PR to fix that.

#306 was merged, so I rebased and switched to `rstrip` 👍

PR #311, opened 2019-01-21
Update README section about generator template

budu:
The correct path use the generator name which isn't pluralized.


composerinteralia:
Thanks!

PR #318, opened 2019-02-06
Use compatible sqlite3 version in test application

y-yagi:
Currently, the `sqlite3` in `Gemfile` generated by `rails new` is not compatible with Rails.

To solve this issue, modified it to use compatible `sqlite3` using template.
Ref: https://github.com/rails/rails/issues/35161


composerinteralia:
Thanks!

Issue #313, opened 2019-02-01
Undefined method 'password' in 'admin' factory


Naveen7696:
Hello,
I tried factory_bot_rails latest version  on a project with Rails 5.0.7 app 
Getting this error:
..
NoMethodError: undefined method 'password' in 'admin' factory
       /app/vendor/bundle/ruby/2.4.0/gems/factory_bot-5.0.0/lib/factory_bot/definition_proxy.rb:97:in `method_missing'
       /app/spec/factories/admins.rb:4:in `block (2 levels) in <top (required)>'
       /app/vendor/bundle/ruby/2.4.0/gems/factory_bot-5.0.0/lib/factory_bot/syntax/default.rb:18:in `instance_eval'
       /app/vendor/bundle/ruby/2.4.0/gems/factory_bot-5.0.0/lib/factory_bot/syntax/default.rb:18:in `factory'
..

after switching back to 4.8.2 the problem goes away...

composerinteralia:
I am guessing it is a result of static attributes being deprecated. Can you try upgrading to 4.11.1 first to see if you get any deprecation warnings?

gnclmorais:
Ah, the Rubocop is for rspec, not minitest… 😭

composerinteralia:
@gnclmorais It is officially only for RSpec, but others have made it work with minitest: https://github.com/thoughtbot/factory_bot/issues/1173#issuecomment-421410343

gnclmorais:
Wow, I was not expecting getting an answer — specially such a useful one! Thank you, @composerinteralia! 🙇🏻‍♂️

KurtRMueller:
I think a more informative error message would be helpful in this instance.

composerinteralia:
I merged https://github.com/thoughtbot/factory_bot/pull/1287 to add a “Did you mean?” style suggestion to the error. I will try to get that released next week.

KurtRMueller:
Thank you for the work you do!

promisepreston:
I think a more explicit answer would have greatly helped.

From the blog by Thoughbot, I came to realize that statically defined attributes have been deprecated from  factory_bot >= 5.0. And so rather than use statically defined attributes for your factories, use dynamically defined attributes

Here's an example

**Use Dynamically defined attribues**
`factory :robot do`
  `name { "Ralph" }`
  `email { "ralph@example.com" }`
`end`

**And not statically defined attributes**

`factory :robot do`
  `name "Ralph"`
[...]

PR #320, opened 2019-02-14
Test against rails 6.0

composerinteralia:


danielricecodes:
For what its worth, I am test upgrading a Rails 5.2 app with `factory_bot_rails` v 4.10 and it appears to work just fine with Rails 6.0.0.beta3.  I actually downgraded my app from FBR 5 to FBR 4 and now my RSpec test suite is 100% green again.

composerinteralia:
@danielricecodes thanks! I had sort of forgotten about this branch.

One reason you may have had to downgrade from FBR 5 to FBR 4 is because we removed static attributes: https://github.com/thoughtbot/factory_bot_rails/issues/313

PR #326, opened 2019-03-18
chore: update readme misspelling

johncblandii:
(moving quickly so feel free to ping me with PR etiquette needs)


composerinteralia:
Thanks! I misspell "definition" in that way far too often...

Issue #323, opened 2019-03-04
new reload behavior breaks normal code reloading with Spring


jakeonfire:
Whenever I change code inside the Rails app/ dir, FactoryBot no longer reloads (per the decisions in this commit: https://github.com/thoughtbot/factory_bot_rails/commit/02a0f584876ff5c7a6d6fa4965da58ed70464a73). This causes a myriad of strange errors, like having multiple class objects for models for which there exist factories (in other words, `MyModel.object_id`, and `my_model_factory_instance.class.object_id` drift apart somehow), ActiveModel methods seem to be forgotten (like `ActiveRecord#store_accessor` not defined errors), and other weirdness.

Adding back
```ruby
if defined?(Spring)
  Spring.after_fork { FactoryBot.reload }
end
```
fixes this issue (for example, in a `config.after_initalize` block in config/environments/test.rb). But this doesn't feel right at all.

I'm not sure if this has something to do with how our app is configured, but it makes using `spring` and `factory_bot_rails` 5.x (and Rails 5.2?) together a non-option (without the workaround mentioned above).

composerinteralia:
I had a feeling this was going to be a problem when I saw https://github.com/thoughtbot/factory_bot/issues/1029#issuecomment-463413328.

We may need to swap out ` reloader.execute_if_updated` in https://github.com/thoughtbot/factory_bot_rails/blob/e1ab9a05686cb52219e38f830f5088379da59276/lib/factory_bot_rails/reloader.rb#L34 with ` reloader.execute`, so any time anything changes in the application factory_bot will reload too. This matches what the rails routes reloader does: https://github.com/rails/rails/blob/2420c44d4b1dfbffea7b624cb8a8c68beb05936f/railties/lib/rails/application/finisher.rb#L144-L154

Thanks for the issue!

jakeonfire:
I had the same thought, but doing so seems to conflict with some other aspect of the loading architecture, yielding (for me):
Factory already registered: adapter (FactoryBot::DuplicateDefinitionError)
...which is the first factory alphabetically.
I'm sure someone can figure it out though.

jakeonfire:
https://github.com/thoughtbot/factory_bot_rails/blob/e1ab9a05686cb52219e38f830f5088379da59276/lib/factory_bot_rails/railtie.rb#L25

composerinteralia:
[...]

PR #329, opened 2019-03-26
Always execute a reload when Rails triggers one (instead of just on u…

jakeonfire:
…pdate) to prevent doubly-defined constants errors in factories when using Spring

- https://github.com/thoughtbot/factory_bot_rails/issues/323

### DIFF
--- a/lib/factory_bot_rails/railtie.rb
+++ b/lib/factory_bot_rails/railtie.rb
@@ -22,10 +22,6 @@ module FactoryBotRails
       Reloader.new(app, config).run
     end
 
-    config.after_initialize do
-      FactoryBot.find_definitions
-    end
-
     private
 
     def definition_file_paths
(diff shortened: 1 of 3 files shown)


jakeonfire:
there are 3 failing specs where `FactoryBot.reload` gets called twice instead of once. i'm not sure why that is though.

composerinteralia:
Maybe because the `to_prepare` hook gets invoke a variable number of times: https://github.com/rails/rails/issues/28108

jakeonfire:
given this comment https://github.com/rails/rails/pull/28111#issuecomment-285370954 should the specs assert that `#reload` is called at least once?

composerinteralia:
Yeah, that makes sense to me

jakeonfire:
@composerinteralia if you feel like adding that spec to prove Spring and `#execute_if_updated` are incompatible, feel free!

composerinteralia:
Thanks! I'll try to find some time on Friday to look at this.

composerinteralia:
I added a failing test in https://github.com/thoughtbot/factory_bot_rails/pull/330

Can you give that a review when you have a chance? It looks like it does indeed pass with your PR. 🎉

composerinteralia:
Thanks again!

PR #330, opened 2019-04-04
Write a failing test for reloading with Spring

composerinteralia:
This test captures what I think is going on in
https://github.com/thoughtbot/factory_bot_rails/issues/323

Currently the factory_bot_rails reloader only reloads factory
definitions when the definition files have changes (using
`execute_if_updated`).

This can cause problems when editing an application file without also
editing a factory_bot definition file. Rails will reload the entire
application, but will not reload the definitions. This causes
factory_bot to have references to classes that were unloaded from the
Rails application and replaced with new classes.

https://github.com/thoughtbot/factory_bot_rails/pull/329 replaces
`execute_if_updated` with `execute` so that factory_bot_rails will
reload the definition files whenever anything in the application is
changed. Hopefully that PR will get this test to pass!

### DIFF
--- a/features/reloading.feature
+++ b/features/reloading.feature
@@ -1,0 +1,63 @@
+Feature:
+  When using factory_bot_rails together with Spring
+  I want changes to my application to trigger the factory_bot_rails reloader
+  So that factory_bot_rails doesn't hold onto stale class references
+
+  Scenario: Editing a model without editing the factory
+    When I successfully run `bundle exec rails new testapp -m ../../features/support/rails_template`
+    And I cd to "testapp"
+    And I add "factory_bot_rails" from this project as a dependency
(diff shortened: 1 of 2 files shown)


iwz:
Chatted IRL: the build was failing for the wrong reasons on previous versions of rails (`bin/rails test` was not available). Daniel switched it to `rake test`

composerinteralia:
Oh no! Now all the tests are passing. That was not supposed to happen.

mike-burns:
Seems right, and the CI failures all seem correct. It seems to pass on Rails 6 -- so this is fixed there, too?

composerinteralia:
I was actually surprised to see the Rails 6 tests passing, but I assume it is because of the new [Zietwerk](https://github.com/fxn/zeitwerk) loader

PR #331, opened 2019-04-04
Do not register a reloader when file paths not exist

### LINKED ISSUE #328: Reloading Issue Triggering Resolved Rails Bug
Related Rails issue:  https://github.com/rails/rails/issues/32700

### Steps to reproduce

- Creating a new rails project with ruby 2.5.3 and rails 6.0.0.beta3
- Add `gem 'factory_bot_rails', '~> 5.0', '>= 5.0.1'` to the gemfile
- Install gems `bundle install`
- boot up the rails server `bundle exec rails s`
- view errors in server output

### Expected behavior
Adding `factory_bot_rails` gem to a rails project should not trigger `** ERROR: directory is already being watched! **` when booting up a rails server or console

### Actual behavior
After including the gem in my project, I'm getting the following errors in my server when running the rails server...
```
be rails s
=> Booting Puma
=> Rails 6.0.0.beta3 application starting in development 
=> Run `rails server --help` for more startup options
        ** ERROR: directory is already being watched! **

        Directory: /path/to/my/app/node_modules/.bin/compression-webpack-plugin

        is already being watched through: /path/to/my/app/node_modules/compression-webpack-plugin

        MORE INFO: https://github.com/guard/listen/wiki/Duplicate-directory-errors
        ** ERROR: directory is already being watched! **

        Directory: /path/to/my/app/node_modules/.bin/mini-css-extract-plugin

        is already being watched through: /path/to/my/app/node_modules/mini-css-extract-plugin

        MORE INFO: https://github.com/guard/listen/wiki/Duplicate-directory-errors
(node:33074) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
        ** ERROR: directory is already being watched! **

        Directory: /path/to/my/app/node_modules/@rails/webpacker/node_modules/.bin/compression-webpack-plugin

        is already being watched through: /path/to/my/app/node_modules/compression-webpack-plugin

        MORE INFO: https://github.com/guard/listen/wiki/Duplicate-directory-errors
        ** ERROR: directory is already being watched! **

        Directory: /path/to/my/app/node_modules/@rails/webpacker/node_modules/.bin/mini-css-extract-plugin

        is already being watched through: /path/to/my/app/node_modules/mini-css-extract-plugin

        MORE INFO: https://github.com/guard/listen/wiki/Duplicate-directory-errors
Puma starting in single mode...
* Version 3.12.1 (ruby 2.5.3-p105), codename: Llamas in Pajamas
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://localhost:3000
```

### System configuration
**Rails version**:  6.0.0.beta3

**Ruby version**:  2.5.3

y-yagi:
If all the paths specified in `FactoryBot.definition_file_paths` do not exist, an empty array is passed to the reloader.

If an empty array is passed, reloader passes it to `Listen.to` as it is. This causes the application root to be watched and `node_modules` to become a target of listening.

The behavior when an empty array is passed to reloader depends on the Rails side. Fixes to not register a reloader when file paths do not exist consistent behavior regardless of Rails side.

Fixes #328.


lib/factory_bot_rails/definition_file_paths.rb
```diff
@@ -19,5 +19,9 @@ def directories
     def files
       @files.select { |file| File.exist?(file) }
     end
+
+    def any?
+      directories.any?
```
composerinteralia: I think this will need to be `directories.any? || files.any?`. 

The case I am thinking of can be tested by adding `spec/fixtures/test/factories.rb` and

```rb
    context "a file exists but not a directory" do
      it "registers a reloader" do
        allow(reloader_class).to receive(:new)

        run_reloader(["spec/fixtures/test/factories", "not_exist_directory"])

        expect(reloader_class).to have_received(:new)
      end
    end
```

y-yagi: Thanks! I fixed.

composerinteralia:
I think we can get CI passing with this change: https://github.com/thoughtbot/factory_bot_rails/pull/332. Let me know what you think.

y-yagi:
Thank you for the fix it! It looks good to me :+1:  
I'm a little concerned about passing that test in my local without that fix.
 
``` 
$ BUNDLE_GEMFILE=gemfiles/rails5.2.gemfile be cucumber features/load_definitions.feature:118
Feature: automatically load factory definitions

  Background:                                                                                        # features/load_definitions.feature:3
    When I successfully run `bundle exec rails new testapp -m ../../features/support/rails_template` # aruba-0.14.9/lib/aruba/cucumber/command.rb:27
    And I cd to "testapp"                                                                            # aruba-0.14.9/lib/aruba/cucumber/file.rb:77
    And I add "factory_bot_rails" from this project as a dependency                                  # features/step_definitions/rails_steps.rb:1
    And I add "test-unit" as a dependency                                                            # features/step_definitions/rails_steps.rb:5
    And I run `bundle install` with a clean environment                                              # features/step_definitions/appraisal.rb:1
    And I write to "db/migrate/1_create_users.rb" with:                                              # aruba-0.14.9/lib/aruba/cucumber/file.rb:53
      """
      migration_class =
        if ActiveRecord::Migration.respond_to?(:[])
          ActiveRecord::Migration[4.2]
        else
          ActiveRecord::Migration
        end

      class CreateUsers < migration_class
        def self.up
          create_table :users do |t|
            t.string :name
          end
        end
      end
      """
    When I run `bundle exec rake db:migrate` with a clean environment                                # features/step_definitions/appraisal.rb:1
    And I write to "app/models/user.rb" with:                                                        # aruba-0.14.9/lib/aruba/cucumber/file.rb:53
      """
      class User < ActiveRecord::Base
      end
      """

  Scenario: use 3rd-party factories with an initializer                                              # features/load_definitions.feature:118
    When I append to "config/application.rb" with:                                                   # aruba-0.14.9/lib/aruba/cucumber/file.rb:61
      """
        require File.expand_path('../../lib/some_railtie/railties.rb', __FILE__)
      """
    When I write to "lib/some_railtie/railties.rb" with:                                             # aruba-0.14.9/lib/aruba/cucumber/file.rb:53
      """
      module SomeRailtie
        class Railtie < ::Rails::Engine
          initializer "some_railtie.factories", :after => "factory_bot.set_factory_paths" do
            FactoryBot.definition_file_paths << File.expand_path('../factories', __FILE__)
          end
        end
      end
      """
    When I write to "lib/some_railtie/factories.rb" with:                                            # aruba-0.14.9/lib/aruba/cucumber/file.rb:53
      """
      FactoryBot.define do
        factory :factory_from_some_railtie, class: 'User' do
          name { 'Artem' }
        end
      end
      """
    When I write to "test/unit/user_test.rb" with:                                                   # aruba-0.14.9/lib/aruba/cucumber/file.rb:53
      """
      require 'test_helper'

      class UserTest < ActiveSupport::TestCase
        test "use factory of some_railtie" do
          user = FactoryBot.create(:factory_from_some_railtie)
          assert_equal 'Artem', user.name
        end
      end
      """
    When I run `bundle exec rake test` with a clean environment                                      # features/step_definitions/appraisal.rb:1
    Then the output should contain "1 assertions, 0 failures, 0 errors"                              # aruba-0.14.9/lib/aruba/cucumber/command.rb:159

1 scenario (1 passed)
14 steps (14 passed)
0m20.056s
``` 

(Does something different between CI and local?)

composerinteralia:
Hm, that is unexpected. It fails for me locally.

y-yagi:
Oh, sorry. My local repo seems to be old. The fail reproduced when rebasing with the latest.
(The `FactoryBot.find_definitions` deleted in https://github.com/thoughtbot/factory_bot_rails/commit/dcfc9f666239be71a56f0ba4f1ad003d3c8550bf remained, and the test passed because of the influence.)

PR #334, opened 2019-04-12
Add back loading definitions in after_initialize

composerinteralia:
We removed the call to `FactoryBot.load_definitions` in the
`after_initialize` hook in dcfc9f6 because the other changes in that
commit were causing us to sometimes call `FactoryBot.load_definitions`
after `FactoryBot.reload`.

We could have changed `FactoryBot.load_definitions` to
`FactoryBot.reload`, which is what this PR does, but it didn't seem
necessary at the time since all the tests were still passing.

We fixed a problem in 9b83f48 that was causing us to watch the root
directory of the project. This fix caused a test failure, but I ended up
changing the test in 2b7bca0 because it seemed like an unlikely
scenario.

This PR adds back the original test case. Although I do think it
unlikely, I would rather not risk making a breaking change in the next
patch release.

### DIFF
--- a/lib/factory_bot_rails/railtie.rb
+++ b/lib/factory_bot_rails/railtie.rb
@@ -22,6 +22,10 @@ module FactoryBotRails
       Reloader.new(app, config).run
     end
 
+    config.after_initialize do
+      FactoryBot.reload
+    end
+
     private
 
     def definition_file_paths
--- a/features/load_definitions.feature
+++ b/features/load_definitions.feature
@@ -115,7 +115,44 @@ Feature: automatically load factory definitions
     When I run `bundle exec rake test` with a clean environment
     Then the output should contain "1 assertions, 0 failures, 0 errors"
 
-  Scenario: use 3rd-party factories with an initializer
+  Scenario: use 3rd-party factories with an initializer and without any user-defined factories
+    When I append to "config/application.rb" with:
+      """
+        require File.expand_path('../../lib/some_railtie/railties.rb', __FILE__)
+      """
+    When I write to "lib/some_railtie/railties.rb" with:
+      """
+      module SomeRailtie
+        class Railtie < ::Rails::Engine
+          initializer "some_railtie.factories", :after => "factory_bot.set_factory_paths" do
+            FactoryBot.definition_file_paths << File.expand_path('../factories', __FILE__)
+          end
+        end
+      end
+      """
+    When I write to "lib/some_railtie/factories.rb" with:
+      """
+      FactoryBot.define do
+        factory :factory_from_some_railtie, class: 'User' do
+          name { 'Artem' }
+        end
+      end
+      """
+    When I write to "test/unit/user_test.rb" with:
+      """
+      require 'test_helper'
+
+      class UserTest < ActiveSupport::TestCase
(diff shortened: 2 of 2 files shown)


lib/factory_bot_rails/railtie.rb
```diff
@@ -22,6 +22,10 @@ class Railtie < Rails::Railtie
       Reloader.new(app, config).run
     end
 
+    config.after_initialize do
+      FactoryBot.reload
```
aledustet: The previous version here was doing:
```ruby 
FactoryBot.find_definitions
```
This new version by doing reload also does:
```ruby
Internal.reset_configuration
register_default_strategies
register_default_callbacks
find_definitions
```
Maybe this question is completely out of context but do we need to re-register all the callbacks and strategies upon initialization? Aren't we doing that somewhere else already?

composerinteralia: Good question. The strategies and callbacks are stored on the configuration object, so `Internal.reset_configuration` wipes them out and we have to re-register them.

composerinteralia:
> The test suite seems to have the entire setup for adding new examples and factories wth a new engine. I was wondering if we should write the failing test when no definition_file_paths are defined an error message suggesting the solution shows?

I am not totally following this question.

aledustet:
> > The test suite seems to have the entire setup for adding new examples and factories wth a new engine. I was wondering if we should write the failing test when no definition_file_paths are defined an error message suggesting the solution shows?
> 
> I am not totally following this question.

What I mean is that the description of the problem on the linked issues was also pointing out that if someone does not changes the engine should also define the definition_file_paths. My question is regarding if we should write a test that mimics this behavior and maybe show an error message hat guides the user in this direction?

composerinteralia:
Hm, I'm sorry I'm still not really following. What would the steps of that test look like?

aledustet:
Maybe is all a misconception but based on [this](https://github.com/thoughtbot/factory_bot_rails/commit/2b7bca00941c20389754d07a362712a2b55d19a7) I'm reading:
```
I am changing this test because there scenario is unlike. If somebody is
using factory_bot definitions from an engine, they will most likely have
their own factory definitions as well, defined in one of the
`config.factory_bot.definition_file_paths`.
```

What if there are deifnitions from an engine and no configuration for `config.factory_bot.definition_file_paths`. Should we show a specific error? That would be the test I was talking about. Making a test case that uses factory bots definitions from an engine without configuring the definitions file path.

composerinteralia:
Thanks for clarifying. I don't think we need anything like that because `config.factory_bot.definition_file_paths` defaults to `FactoryBot.definition_file_paths` (see [here](https://github.com/thoughtbot/factory_bot_rails/blob/master/lib/factory_bot_rails/railtie.rb#L11)), which includes all the usual locations for defining factory definitions.

So the scenarios I wanted to cover here are:

1. Engine includes factories using an initializer, & there are no user-defined factories
2. Engine includes factories using an initializer, & there are some user-defined factories
3. Engine includes factories by appending to `config.factory_bot.definition_file_paths` (in this case it doesn't make any difference whether there are user-defined factories)

We will now have tests for all three of these scenarios.

PR #339, opened 2019-05-15
Fix definition_file_paths setting for Rails engine

mtsmfm:
### 1. Add `require`

Even if both Rails engine gem and factory_bot_rails gem are written on Gemfile, `defined?(FactoryBotRails)` returns false because it depends on the order on Gemfile.

### 2. `definition_file_paths` is an Array

I'm not sure it should be `+=` or `=` though, anyway, it must be an Array.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,14 @@ This will cause factory\_bot\_rails to automatically load factories in
 It is possible to use this setting to share factories from a gem:
 
 ```rb
+begin
+  require 'factory_bot_rails'
+rescue LoadError
+end
+
 class MyEngine < ::Rails::Engine
   config.factory_bot.definition_file_paths =
-    File.expand_path('../factories', __FILE__) if defined?(FactoryBotRails)
+    [File.expand_path('../factories', __FILE__)] if defined?(FactoryBotRails)
 end
 ```
 


composerinteralia:
Thanks for the PR.

## 1 I am not used to manually requiring gems in a Rails project, since normally the `config/application.rb` loads all the gems at once with `Bundle.require(*Rails.groups)`. Let's skip that change for this PR.

## 2 Yes, that definitely needs to be an array. Good catch! Let's use `+=`

mtsmfm:
> 1 - I am not used to manually requiring gems in a Rails project, since normally the `config/application.rb` loads all the gems at once with `Bundle.require(*Rails.groups)`. Let's skip that change for this PR.

Hmm, I think we should mention in docs about the order of `require` because it does NOT change `definition_file_paths`. 

### NG

```ruby
...
gem "rails"
gem "my_engine"

group :test do
  gem "factory_bot_rails"
end

# `defined? (FactoryBotRails)` returns `false` because my_engine is required before loading factory_bot_rails
```

### OK

```ruby
...
gem "rails"

group :test do
  gem "factory_bot_rails"
end

gem "my_engine"
```

If we just remove `defined?(FactoryBotRails)`, `config.factory_bot.definition_file_paths` raises `NoMethodError` because factory_bot_rails is initialized here:

https://github.com/thoughtbot/factory_bot_rails/blob/c1d0a7887beb2418200e4171642341577d3832b3/lib/factory_bot_rails/railtie.rb#L10

As another option, we can use `before_initialize`:

```ruby
class MyEngine < ::Rails::Engine
  config.before_initialize do
    config.factory_bot.definition_file_paths += [MyEngine.root.join("spec/factories")] if defined?(FactoryBotRails)
  end
end
```

composerinteralia:
Thanks for that explanation. If the require with rescue solves the problem, then let's keep include that in the documentation as you have done here. 👍

PR #343, opened 2019-07-26
Set up reloading after_initialize

### LINKED ISSUE #336: Issues with custom error message translations
We have a model class in rails with the following validation:

```ruby
  validates :items, length: {
    minimum: 1, message: I18n.t('items.one_required')
  }
```

This works fine in version 5.0.1, but in 5.0.2 there's a `translation missing: en.items.one_required ` in the errors collection.  Not sure why factory_bot would affect the i18n class but it appears to.

composerinteralia:
Fixes #336

Before this commit, the initialization process looked like
this:

1. Run the ["factory_bot.set_factory_paths"][set_factory_paths] initializer
2. Run the ["factory_bot.register_reloader"][register_reloader] initializer, which sets up a [prepare callback][]
4. This triggers the factory_bot [prepare callback][], which causes factory\_bot to [reload][]

1. [I18n initializes]

The double reloading of factory_bot in this initialization is not ideal,
but also shouldn't generally cause any problems on its own.

The problems people are having in #336 come from the fact that
I18n gets set up in an `after_initialize` callback, but factory_bot gets
reloaded before the `after_initialize` callbacks are triggered.
If the `FactoryBot.define` block references any code that uses I18n
translations as it loads, that code will raise an error (references
inside other factory_bot methods, or code that uses I18n translations
inside of methods still works fine, since the whole Rails initialization
process would be complete by the time any of that code runs).

This commit moves factory_bot reloading from a prepare callback into an
`after_initialize` callback. This allows us to avoid reloading
factory_bot before I18n is loaded, and also gets rid of that pesky
double reloading of factory_bot.

The new initialization process looks like:

2. Run the [`:run_prepare_callbacks`][] initializer, which no longer involves factory_bot

Rails runs `after_intialize` callbacks
1. [I18n initializes]
2. Run the factory_bot reloader, which sets up the prepare callback for
any future prepares (for example calling `reload!` in the console), and
executes the reloader to run the initial `FactoryBot.reload`

### DIFF
--- a/lib/factory_bot_rails/reloader.rb
+++ b/lib/factory_bot_rails/reloader.rb
@@ -4,21 +4,22 @@ require "factory_bot_rails/definition_file_paths"
 
 module FactoryBotRails
   class Reloader
-    def initialize(app, config)
+    def initialize(app)
       @app = app
-      @config = config
       @paths = DefinitionFilePaths.new(FactoryBot.definition_file_paths)
     end
 
     def run
       return unless @paths.any?
 
-      register_reloader(build_reloader)
+      reloader = build_reloader
+      register_reloader(reloader)
+      reloader.execute
     end
 
     private
 
-    attr_reader :app, :config
+    attr_reader :app
 
     def build_reloader
       reloader_class.new(@paths.files, @paths.directories) do
@@ -31,11 +32,11 @@ module FactoryBotRails
     end
 
     def register_reloader(reloader)
-      config.to_prepare do
+      app.reloaders << reloader
+
+      app.reloader.to_prepare do
         reloader.execute
       end
-
-      app.reloaders << reloader
     end
   end
 end
--- a/lib/factory_bot_rails/railtie.rb
+++ b/lib/factory_bot_rails/railtie.rb
@@ -18,12 +18,8 @@ module FactoryBotRails
       FactoryBot.definition_file_paths = definition_file_paths
     end
 
-    initializer "factory_bot.register_reloader" do |app|
-      Reloader.new(app, config).run
-    end
-
-    config.after_initialize do
-      FactoryBot.reload
+    config.after_initialize do |app|
+      Reloader.new(app).run
     end
 
     private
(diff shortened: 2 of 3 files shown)


composerinteralia:
I would like to get a failing test before merging this, but I had trouble making that happen.

@y-yagi it would be great if you could take a look at this PR, if you have a moment. I seem to be playing Whac-A-Mole with the bugs related to this reloading feature.

lib/factory_bot_rails/reloader.rb
```diff
@@ -4,21 +4,22 @@
 
 module FactoryBotRails
   class Reloader
-    def initialize(app, config)
+    def initialize(app)
```
aledustet: Is there anything we should include in the docs now there is not a config to be passed in?

lib/factory_bot_rails/reloader.rb
```diff
@@ -31,11 +32,11 @@ def reloader_class
     end
 
     def register_reloader(reloader)
-      config.to_prepare do
+      app.reloaders << reloader
+
+      app.reloader.to_prepare do
```
aledustet: The way I understand this, the ```to_prepare``` block that was part of the config before now will be under the reloaders?

lib/factory_bot_rails/reloader.rb
composerinteralia: This class is not meant for public use, so changing the API should be fine.

lib/factory_bot_rails/reloader.rb
composerinteralia: In later versions of Rails this amounts to the same thing. `config.to_prepare` [pushes the callbacks into an array](https://github.com/rails/rails/blob/master/railties/lib/rails/railtie/configuration.rb#L79-L83) which later gets [copied over to `app.reloader.to_prepare`](https://github.com/rails/rails/blob/master/railties/lib/rails/application/finisher.rb#L104-L108).

But it didn't work quite like this in Rails 4.2, which is why we are getting CI failures.

composerinteralia:
I am going to close this for now. When we drop support for Rails 4.2 we can revisit this, and I agree we will definitely want to keep the test from #347.

PR #347, opened 2019-09-20
Ensure factory_bot only loads after initialization

### LINKED ISSUE #336: Issues with custom error message translations

composerinteralia:
Fixes #336
Alternate solution to #343

The initialization process looks like this:

initializer
2. Run the ["factory_bot.register_reloader"][register_reloader]
initializer, which sets up a [prepare callback][]
4. This triggers the factory_bot [prepare callback][], which causes
factory\_bot to [reload][]

1. [I18n initializes]

The double reloading of factory_bot in this initialization is not ideal,
but also shouldn't generally cause any problems on its own.

The problems people are having in #336 come from the fact that
I18n gets set up in an `after_initialize` callback, but factory_bot gets
reloaded before the `after_initialize` callbacks are triggered.
If the `FactoryBot.define` block references any code that uses I18n
translations as it loads, that code will raise an error (references
inside other factory_bot methods, or code that uses I18n translations
inside of methods still works fine, since the whole Rails initialization
process would be complete by the time any of that code runs).

This commit changes step 4 above to avoid reloading factory_bot before
the application has initialized.

Co-authored-by: Danny Garcia <dannygarcia.me@gmail.com>

### DIFF
--- a/features/reloading.feature
+++ b/features/reloading.feature
@@ -63,3 +63,45 @@ Feature:
     And I run `spring stop` with a clean environment
     Then the output should contain "1 runs, 1 assertions"
     And the output should not contain "Failure:"
+
+Scenario: Initializing the reloader with I18n support
+    When I successfully run `bundle exec rails new testapp -m ../../features/support/rails_template`
+    And I cd to "testapp"
+    And I add "factory_bot_rails" from this project as a dependency
+    And I add "test-unit" as a dependency
+    And I run `bundle install` with a clean environment
+    And I run `bundle exec rake db:migrate` with a clean environment
+    And I write to "app/models/user.rb" with:
+      """
+      class User
+        TRANSLATION = I18n.translate("translation_key")
+      end
+      """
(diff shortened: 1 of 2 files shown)


features/reloading.feature
```diff
@@ -63,3 +63,45 @@ Feature:
     And I run `spring stop` with a clean environment
     Then the output should contain "1 runs, 1 assertions"
     And the output should not contain "Failure:"
+
+Scenario: Initializing the reloader with I18n support
```
aledustet: very self-explanatory and specific test 👏

composerinteralia: Thanks for the review! I will merge this and close #343 for now.

PR #353, opened 2019-10-21
Fix rubocop offenses in generated files

nicoalonsop:
When scaffolding a new model the generator generates the factory with rubocop offenses (using default rubocop configuration for rails).

Fixed offenses:
- Style/StringLiterals
- Style/FrozenStringLiteralComment

### DIFF
--- a/features/generators.feature
+++ b/features/generators.feature
@@ -18,7 +18,7 @@ Feature:
       """
       FactoryBot.define do
         factory :user do
-          name { "MyString" }
+          name { 'MyString' }
           age { 1 }
         end
       end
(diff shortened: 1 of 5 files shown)


composerinteralia:
Thanks for the PR! thoughtbot uses https://github.com/thoughtbot/guides/blob/master/style/ruby/.rubocop.yml as the default config (it prefers double quotes and disables the frozen string literal comments), so I am going to close this.

You can read more about our [choice of single quotes](https://github.com/thoughtbot/guides/pull/565), and our [choice of disabling the frozen string literal comment](https://github.com/thoughtbot/guides/pull/446) in the guides repo.

PR #358, opened 2020-01-05
more sofisticated generation

dima4p:


composerinteralia:
Is this different than https://github.com/thoughtbot/factory_bot_rails/pull/340?

dima4p:
I little. The test were added and the bugs fixed.

Gemfile.lock
```diff
@@ -183,4 +183,4 @@ DEPENDENCIES
   uglifier
 
 BUNDLED WITH
-   2.0.2
+   2.1.3
```
entcheva: Can we separate dependency updates from this PR?

factory_bot_rails.gemspec
```diff
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name        = "factory_bot_rails"
-  s.version     = "5.1.1"
+  s.version     = "5.1.1.1"
```
entcheva: Let's remove this change, we'll bump the version when we cut a new release.

features/generators.feature
```diff
@@ -10,16 +10,22 @@ Feature:
...
     And the output should contain "test/factories/namespaced/users.rb"
     And the file "test/factories/users.rb" should contain exactly:
       """
       FactoryBot.define do
         factory :user do
-          name { "MyString" }
-          age { 1 }
+          sequence(:name) {|n| "Name#{format '%03d', n}" }
+          sequence(:age) {|n| "1#{format '%03d', n}" }
+          association :company, factory: :company
```
entcheva: Auto-generating associations sounds like a good improvement.

What are your thoughts on splitting this out into its own PR?

lib/generators/factory_bot/model/model_generator.rb
```diff
@@ -61,11 +61,34 @@ def factory_definition
         RUBY
       end
 
+      # rubocop:disable Metrics/MethodLength, Layout/LineLength, Metrics/PerceivedComplexity
```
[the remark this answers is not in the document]
entcheva: Why are we disabling PerceivedComplexity here? What are your thoughts on working to simplify this code instead of disabling this rule?

Gemfile.lock
dima4p: sure, we just have different environment.

factory_bot_rails.gemspec
dima4p: sure, but I have to have it while my code is not in the master.

features/generators.feature
dima4p: this code is directly connected with the changes in lib/generators/factory_bot/model/model_generator.rb and covers it with tests

dima4p:
To the last question there is no place to answer, so I do it here.

> Why are we disabling PerceivedComplexity here? What are your thoughts on working to simplify this code instead of disabling this rule?

Sometimes formal approach make thing worse. I do not see the way to simplify the code in a way it will be more clear for human.

entcheva:
> Sometimes formal approach make thing worse. I do not see a way to simplify the code in a way it will be more clear for human.

I think there's value in sticking to the Rubocop rule on PerceivedComplexity here. There's a suggestion in #340 on how to simplify your method so we don't have to disable Rubocop:

> Smaller methods with clear names are generally easier than one long method, so you could start refactoring by pulling out smaller helper methods.

features/generators.feature
entcheva: Sure. I suggest pulling this line, the applicable code in `model_generator.rb`, and the corresponding test, and breaking it out into a separate PR.

dima4p:
> > Smaller methods with clear names are generally easier than one long method, so you could start refactoring by pulling out smaller helper methods.

the whole method consists indeed in checking one of 9 conditions and and returning ONE string.
The decomposition it in methods that will check only one condition and then call next method that will do the same and so 9 times will make the bot happy, but not the person, that will read the text.
Also the returning string is long as it is a generator. Will splitting a string make it more clear?

composerinteralia:
Thank you for your work here, @dima4p.

I am going to close this PR. The code we currently have uses the [default attribute values from Rails](https://github.com/rails/rails/blob/0196551e6039ca864d1eee1e01819fcae12c1dc9/railties/lib/rails/generators/generated_attribute.rb#L91-L107), and I think it makes sense to stick to those.

This change makes some assumptions that I don't think are appropriate for every project. Age is a good example of where some of these assumptions don't work. The code here ends up generating the ages "1001", "1002", "1003", etc. While 1001 make make sense for lots of integers, it doesn't really make sense for ages.

I am not comfortable making these assumptions for everyone who uses factory_bot_rails, and would rather stick with the tried-and-true values provided by Rails itself.

I am also in agreement with @entcheva that the Rubocop rule on PerceivedComplexity is meaningful here. The code in this PR adds a lot of additional complexity to the generator, and I don't think it makes sense for us to take on that additional complexity, given that what we currently have works well enough in most cases.

As with fixtures, I don't think it is possible to generate a factory definition that will work for all use cases in all applications. There is always going to be some amount of manual customization. I think of these generators as giving people the rough boilerplate, making it easier for them to focus on the parts that are specific to their application.

dima4p:
> Thank you for your work here, @dima4p.
> 
> I am going to close this PR. The code we currently have uses the [default attribute values from Rails](https://github.com/rails/rails/blob/0196551e6039ca864d1eee1e01819fcae12c1dc9/railties/lib/rails/generators/generated_attribute.rb#L91-L107), and I think it makes sense to stick to those.
> 
> This change makes some assumptions that I don't think are appropriate for every project. Age is a good example of where some of these assumptions don't work. The code here ends up generating the ages "1001", "1002", "1003", etc. While 1001 make make sense for lots of integers, it doesn't really make sense for ages.
> 
> I am not comfortable making these assumptions for everyone who uses factory_bot_rails, and would rather stick with the tried-and-true values provided by Rails itself.
> 
> I am also in agreement with @entcheva that the Rubocop rule on PerceivedComplexity is meaningful here. The code in this PR adds a lot of additional complexity to the generator, and I don't think it makes sense for us to take on that additional complexity, given that what we currently have works well enough in most cases.
> 
> As with fixtures, I don't think it is possible to generate a factory definition that will work for all use cases in all applications. There is always going to be some amount of manual customization. I think of these generators as giving people the rough outline, making it easier for them to focus on the parts that are specific to their application.

I use the code and the analogue of it in factory girl for many years. It in 95% cases does not require any further change and saves 90% time required to make the factories workable. It is often required to distinguish two instances. Also in view specs.
It's a pity other developers will not have it…

composerinteralia:
@dima4p For developers who come across this thread and would like to use this code, I think your gem https://github.com/dima4p/factory_girl_fixtures_template would be a good fit. It may be worth releasing a factory_bot-compatible version of that gem as well (i.e. factory_bot_fixtures_template).

dima4p:
> @dima4p For developers who come across this thread and would like to use this code, I think your gem https://github.com/dima4p/factory_girl_fixtures_template would be a good fit. It may be worth releasing a factory_bot-compatible version of that gem as well (i.e. factory_bot_fixtures_template).

As soon as FactoryBot will use template for the generator I will do that. But at the moment, unfortunately, my patch is the only way to implement it. Of course, I have my own branch for my projects, but I prefer to have it public.

composerinteralia:
I brought back the ability to use custom templates in https://github.com/thoughtbot/factory_bot_rails/pull/307.

dima4p:
> I brought back the ability to use custom templates in #307.

Great news. I will have a look. Thanks.

Issue #356, opened 2019-12-03
Model generation ignores --force-plural


armstnp:
When generating a model using `--force-plural`, the factory_bot adapter still emits the singular form as the factory name.

Example:
```
> rails g model FileMedia filename:string --force-plural
...
invoke      factory_bot
create        spec/factories/file_media.rb
```

This is working as expected. The contents of the factory, however, are:
```ruby
FactoryBot.define do
  factory :file_medium, class: 'FileMedia' do
    filename { "MyString" }
  end
end
```
Notice that the auto-inflected singular form `:file_medium` is used instead of the requested plural, `:file_media`.

composerinteralia:
The problematic line in question gets generated at https://github.com/thoughtbot/factory_bot_rails/blob/ce94dc1a39c642dbb56e8d005f4c6a3ab6053bd5/lib/generators/factory_bot/model/model_generator.rb#L57 if anybody wants to look into making a PR.

We are using the `singular_table_name` method: https://github.com/rails/rails/blob/05a6cd5565a808e1063bc6c7cbdfe5ae6a46b72c/railties/lib/rails/generators/named_base.rb#L113-L115 but maybe there is something else we could be using.

armstnp:
Walking up the inheritance chain, I think the flag may be visible via (`pluralize_table_names?`)[https://api.rubyonrails.org/classes/Rails/Generators/NamedBase.html#method-i-pluralize_table_names-13F] on the model generator object.

entcheva:
If this is fixed, the factory we define would look like

```rb
FactoryBot.define do
  factory :file_media, class: 'FileMedia' do
    filename { "MyString" }
  end
end
```

PR #366, opened 2020-03-28
Fix for factory class naming

alextrueman:
Changing the process of name selection for factory defining

Example:
```
rails g model UserMedia filename:string --force-plural
...
invoke      factory_bot
create        spec/factories/user_media.rb
```
it generates the correct factory naming:
```ruby
FactoryBot.define do
  factory :user_media do
    filename { "MyString" }
  end
end
```

Fixes: [356](https://github.com/thoughtbot/factory_bot_rails/issues/356)


composerinteralia:
Hey there, Alex! It's been a while! Thanks for the contribution.

The build hasn't been passing on master for a while. I am trying to fix that in https://github.com/thoughtbot/factory_bot_rails/pull/367 (still failing for Rails 6 at the moment).

alextrueman:
> Hey there, Alex! It's been a while! Thanks for the contribution.
> 
> The build hasn't been passing on master for a while. I am trying to fix that in #367 (still failing for Rails 6 at the moment).

Hey there Daniel! You are right, it's been a while 😅

Yeah, I was commented your PR about this failing, I'm pretty sure we should try with rails 6.0.0 and cucumber >= 1.8.0

alextrueman:
@composerinteralia hey there, it's green now :) May I ask you for review? 😅

lib/generators/factory_bot.rb
```diff
@@ -14,8 +14,12 @@ def self.source_root
         File.expand_path(path)
       end
 
+      def explicit_class_name
```
composerinteralia:
```suggestion
      def factory_name
```

lib/generators/factory_bot.rb
```diff
@@ -14,8 +14,12 @@ def self.source_root
...
+        class_name.gsub("::", "").underscore
+      end
+
       def explicit_class_option
-        return if class_name == singular_table_name.camelize
+        return if class_name.underscore == explicit_class_name
```
composerinteralia:
```suggestion
        return if class_name.underscore == factory_name
```

lib/generators/factory_bot/model/model_generator.rb
```diff
@@ -54,7 +54,7 @@ def create_factory_file
 
       def factory_definition
         <<~RUBY
-            factory :#{singular_table_name}#{explicit_class_option} do
+            factory :#{explicit_class_name}#{explicit_class_option} do
```
composerinteralia:
```suggestion
            factory :#{factory_name}#{explicit_class_option} do
```

I think this name captures the purpose of the method a bit better. We are using it to get the name of the factory.

alextrueman: @composerinteralia Thank you for the suggestion! 😅

PR #367, opened 2020-03-28
Update rails versions on Travis 

composerinteralia:
This commit gets the Travis build passing by:

* Removing Rails 4.2 so we don't have to deal with bundler < 2 anymore
* Bumping to latests Rails versions
* Explicitly listing all gems that appear in the default Rails Gemfile
* Avoiding bootsnap and javascript, since they aren't relevant to these tests


Appraisals
```diff
@@ -20,28 +15,40 @@ appraise "rails5.1" do
...
-  gem "jbuilder"
-  gem "listen", "~> 3.1.5"
-  gem "puma", "~> 3.11"
-  gem "sass-rails", "~> 5"
+  gem "bootsnap", ">= 1.4.2", require: false
+  gem "byebug", platforms: :ruby
+  gem "capybara", ">= 2.15"
+  gem "jbuilder", "~> 2.7"
+  gem "puma", "~> 4.1"
+  gem "rails", "~> 6.0.2", ">= 6.0.2.2"
```
alextrueman: Regarding [this](https://github.com/cucumber/cucumber-rails/issues/461) issue, looks like it will failing for a while 😞

composerinteralia: 😢 Thanks for looking into this. I'm going to see if I can cucumber to log more details about what is going wrong here.

alextrueman:
:tada: 👍

composerinteralia:
Finally passing! I'm going to try to do a bit of cleanup, and then will ask you for a review @alextrueman. Thanks for your help here.

composerinteralia:
@alextrueman I'm happy with this. If it looks good to you, I will go ahead an merge and then you can rebase on master to get the build passing on your PR.

features/reloading.feature
```diff
@@ -4,8 +4,7 @@ Feature:
   So that factory_bot_rails doesn't hold onto stale class references
 
   Scenario: Editing a model without editing the factory
-    When I successfully run `bundle exec rails new testapp -m ../../features/support/rails_template`
-    And I cd to "testapp"
+    When I create a new rails application
     And I add "factory_bot_rails" from this project as a dependency
     And I add "test-unit" as a dependency
```
alextrueman: Since we remove `gem 'test-unit'` do we need it now?

composerinteralia: Yeah I think you are correct that we can remove this, since test-unit already ships with Ruby.

features/step_definitions/rails_steps.rb
```diff
@@ -1,3 +1,14 @@
+When /^I create a new rails application$/ do
+  options = "--skip-bootsnap --skip-javascript"
+  template = "-m ../../features/support/rails_template"
+  result = run_command("bundle exec rails new test_app #{options} #{template}")
```
alextrueman: What you think about creating an app in API mode? (`.. --api`) it will skip generating views, helpers, and assets. I think it might increase tests speed 🤔

alextrueman: I also think about:
```
--skip-action-mailer
--skip-active-storage
--skip-action-cable
```
for example, I might be wrong, but I think it might be helpful :)

composerinteralia: Yup, I think that is a good idea. The tests already got a bit faster by skipping js. And the fewer dependencies we have the less flaky these tests become (many of our failures are a result of some incompatible version of some gem).

I was also thinking it would be nice if we didn't need to install the development dependencies, since we don't ever start the app in development mode. Perhaps we could also `--skip bundle` and then manually run `bundle install --without development`.

I am going to merge this as is for now to get a passing build on master, and then open a separate issue to explore reducing the number of gems we need to install.

alextrueman: Sounds good let's do it! 👍

composerinteralia: I opened #370

Issue #370, opened 2020-03-29
Install fewer gems on Travis


composerinteralia:
As discussed in https://github.com/thoughtbot/factory_bot_rails/pull/367#discussion_r399810206

Our tests generate a full rails app, but I don't think they need most of the libraries included by default. Unneeded gems make our test suite slower, and increase the chance of getting errors when new versions of each gem come out.

We may not need to include all of these libraries. Passing `--skip javascript`, for example, sped up our test suite significantly.  We may also want to pass `--skip-action-mailer`, `--skip-active-storage`, `--skip-action-cable`, etc. We could also try passing `--api`.

Additionally, we may want to explore passing `--skip bundle` and then manually running `bundle install --without development` to avoid installing unnecessary development gems.

alextrueman:
I think we can split it into a few PRs:
1. With adding pass options(--skip-..., --api)
2. With experiments with `--skip bundle`
3. With the revision of the gems

PR #371, opened 2020-03-29
Add skip options for test app creation

alextrueman:
Related to point 1 of [this](https://github.com/thoughtbot/factory_bot_rails/issues/370#issuecomment-605666776) comment

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -47,15 +47,12 @@ end
 appraise "rails6.0" do
   gem "byebug"
   gem "capybara", ">= 2.15"
-  gem "jbuilder", "~> 2.7"
   gem "listen", ">= 3.0.5", "< 3.2"
   gem "puma", "~> 4.1"
   gem "rails", "~> 6.0.2", ">= 6.0.2.2"
-  gem "sass-rails", ">= 6"
   gem "selenium-webdriver"
   gem "spring"
   gem "spring-watcher-listen", "~> 2.0.0"
   gem "sqlite3", "~> 1.4"
   gem "web-console", ">= 3.3.0", group: :development
-  gem "webdrivers"
 end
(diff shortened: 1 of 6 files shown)


alextrueman:
Hey there @composerinteralia 😅
Looks like it became a bit faster, but I was required to remove some gems in this PR(we can merge 1st and 3rd points of the issue in this PR, I think, what your thoughts on it?) 😅

Appraisals
```diff
@@ -47,15 +47,12 @@ end
 appraise "rails6.0" do
   gem "byebug"
   gem "capybara", ">= 2.15"
-  gem "jbuilder", "~> 2.7"
```
composerinteralia, line 50: Let's remove these gems from the other versions as well, if possible. We might speed things up a little for those test runs as well.

features/step_definitions/rails_steps.rb
```diff
@@ -1,5 +1,14 @@
 When /^I create a new rails application$/ do
-  options = "--skip-bootsnap --skip-javascript"
+  options =
+    %w[
+      --api
```
composerinteralia, line 4: It looks like [`--api` automatically takes care of `--skip-javascript` and `--skip-sprockets`](https://github.com/rails/rails/blob/157920aead96865e3135f496c09ace607d5620dc/railties/lib/rails/generators/rails/app/app_generator.rb#L277-L279), but I sort of like keeping those options here anyway, to make it more explicit what we are trying to leave out. 👍

Appraisals
composerinteralia: Let's also see if we can get rid of capybara, selenium-webdriver, and web-console, which shouldn't be needed anymore since we are in api mode.

composerinteralia:
Excellent! I'm loving this much faster test suite. These test used to 6-12 minutes each, now they are 2-3 minutes each.

alextrueman:
> Excellent! I'm loving this much faster test suite. These test used to 6-12 minutes each, now they are 2-3 minutes each.

Yeah, it's awesome, I wasn't thought it become so much faster :sweat_smile:

PR #372, opened 2020-03-31
Add skipping of bundle install for test app

alextrueman:
Related to point [2](https://github.com/thoughtbot/factory_bot_rails/issues/370#issuecomment-605666776) of this comment


composerinteralia:
Oh, neat. I imagined we would need to to run `bundle install --without development` at some point, but I guess we don't need to do that since the gems are already installed and we already have the Gemfile.lock available.

PR #374, opened 2020-04-04
Test against ruby 2.7

composerinteralia:
And drop support for EOL 2.3 (holding off on dropping 2.4 since the
website hasn't yet been updated to show that it is EOL)

https://www.ruby-lang.org/en/downloads/branches/


alextrueman:
🚀

PR #388, opened 2020-12-10
Test against Rails 6.1

berkos:
👋    Test against Rails 6.1


composerinteralia:
Damn, that was fast. Thanks!

berkos:
> Damn, that was fast. Thanks!

No Problem! It looks like all the builds are failing so trying to find which gem might have been updated and is causing the issue. 🤔

jpac-run:
Could be related: https://github.com/rails/rails/issues/40871

composerinteralia:
@berkos it looks like it was a problem with spring 2.1.1 (see https://github.com/thoughtbot/factory_bot_rails/pull/389). Should be fixed now, so can you rebase this branch on the latest commit to see if it passes?

Appraisals
```diff
@@ -38,3 +38,13 @@ appraise "rails6.0" do
   gem "spring-watcher-listen", "~> 2.0.0"
   gem "sqlite3", "~> 1.4"
 end
+
+appraise "rails6.1" do
+  gem "byebug"
+  gem "listen", "~> 3.2"
+  gem "puma", "~> 4.1"
```
composerinteralia, line 45: Looks like the [default Puma version](https://github.com/rails/rails/blob/v6.1.1/railties/lib/rails/generators/app_base.rb#L209) for a new Rails app is "~> 5.0", so we will probably need to match that to get the tests passing (basically we want the versions here in the appraisal to match whatever is in the Gemfile of a brand new Rails app).

```suggestion
  gem "puma", "~> 5.0"
```

berkos: Makes sense, thanks for the review and spotting this! Resolved.

Appraisals
```diff
@@ -38,3 +38,13 @@ appraise "rails6.0" do
   gem "spring-watcher-listen", "~> 2.0.0"
   gem "sqlite3", "~> 1.4"
 end
+
+appraise "rails6.1" do
+  gem "byebug"
+  gem "listen", "~> 3.2"
+  gem "puma", "~> 5.0"
+  gem "rails", "~> 6.1.0", ">= 6.1.0.0"
+  gem "spring"
```
composerinteralia, line 47: Ah, sorry I missed this in the last review, we will need to avoid the broken version of spring here too


```suggestion
  gem "spring", "!= 2.1.1"
```

gemfiles/rails6.1.gemfile
```diff
@@ -0,0 +1,19 @@
...
+gem "aruba"
+gem "cucumber", "1.3.19"
+gem "rake"
+gem "rspec-rails"
+gem "standard"
+gem "byebug"
+gem "listen", "~> 3.2"
+gem "puma", "~> 5.0"
+gem "rails", "~> 6.1.0", ">= 6.1.0.0"
+gem "spring"
```
composerinteralia, line 15:
```suggestion
gem "spring", "!= 2.1.1"
```

Appraisals
composerinteralia: I'll go ahead and merge so we can get the tests running

PR #389, opened 2021-01-08
Skip Spring version 2.1.1

entcheva:
Changes introduced in Spring 2.1.1 (rails/spring#621) are breaking some tests. That change was reverted in rails/spring#629 but hasn't been released yet. Until rails/spring#629 is released, this PR skips Spring version 2.1.1.

Co-authored-by: Daniel J. Colson <daniel.colson@hey.com>


composerinteralia:
I occurred to me we saw some recent failures in suspenders that are probably similar to this. For some reason commands that used to run in a test app directory started getting run in the top-level suspenders directory. I think that is what is happening here: the test is in factory_bot_rails/tmp/aruba/test_app/test, but for some reason something somewhere is looking for it in factory_bot_rails/.

I don't think this is a good solution, but if we change `bin/rails test` to `bin/rails test tmp/aruba/test_app/test/unit/user_test.rb` it passes for me locally. Interestingly the `bin/rails` part works find even though that is also in tmp/aruba/test_app.

I would love to know what library changed this behavior for us, but not sure how to find out. We were having some rake issues, so maybe something in there somewhere?

composerinteralia:
Figured it out. It was introduced in spring 2.1.1 (https://github.com/rails/spring/pull/621). That change was reverted in https://github.com/rails/spring/pull/629, but hasn't been released yet. So I think the solution here is to avoid the broken version of spring. I think we can do that by adding a "!= 2.1.1 constraint to the Gemfile", updating the appraisals, and adding `"gsub_file "Gemfile", /^  gem 'spring'$/, '  gem "spring", "!= 2.1.1"'` to features/support/rails_template .

cc @garciadanny and @mike-burns in case this helps you with the suspenders tests. Or if you wait long enough they might fix themselves when a new version of spring comes out.

features/support/rails_template
```diff
@@ -1,3 +1,4 @@
 if Rails.gem_version < Gem::Version.new('6')
   gsub_file "Gemfile", /^gem 'sqlite3'$/, 'gem "sqlite3", "~> 1.3.6"'
+  gsub_file "Gemfile", /^ gem 'spring'$/, ' gem "spring", "!= 2.1.1"'
```
composerinteralia, line 3: I think we want this outside the `if Rails.gem_version < Gem::Version.new('6')` check, since it is failing in every version.

entcheva: Good call! Updated in c301ff9.

I still see one local failing `reloading.feature` test that times out. Is it green for you? Let's see what CI has to say.

composerinteralia: Although the rails 6 tests passed anyway so 🤷🏻‍♂️

composerinteralia: Pushed up an additional commit to fix the standard style violation.

composerinteralia:
BTW, this is the Travis announcement I was thinking of, specifically the part about OSS: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

features/support/rails_template
entcheva: Thank you! We have a green test suite, should we squash & merge?

composerinteralia: 🎉 Yippee! I'll leave it to you.

composerinteralia:
@entcheva it seems like we may want to [migrate from Travis to Circle](https://schneems.com/2021/01/13/migrating-a-ruby-library-from-travisci-to-circleci/) at some point.

entcheva:
@composerinteralia Nice! I especially appreciate the performance comparison to Github Actions. This is a helpful breakdown, thank you!

PR #396, opened 2021-05-26
Replace Travis with GitHub Actions

### LINKED ISSUE #395: Switch from Travis CI to GitHub Actions
We've already done this on factory_bot, and it would be nice for them to match. We could probably copy over the config from there and then adjust as needed.

composerinteralia:
Pulled from https://github.com/thoughtbot/factory_bot/blob/23162d4348ed8b316d36091b0fc734c174be75c3/.github/workflows/build.yml

As with factory_bot, we're no longer running against Ruby head, since we
don't have a great way to do allowed failures.

Closes #395


composerinteralia:
Looks like we're almost there, but getting an error from spring that it's not able to find the application config:

`lib/spring/configuration.rb:31:in `application_root_path': Spring::MissingApplication (Spring::MissingApplication)`

mathieujobin:
as far as I can see, there is only a couple of tests failing related to Spring.
I don't personally use Spring, so I could not care less. can we skip those two? and move forward until someone has the time to investigate? 

I'd like to see this merged first and then rails 7 added to the mix to ensure its fully supported.

thanks

composerinteralia:
I agree it'd be nice to get CI working against Rails 7.

In my experience skipping tests is basically the same as deleting them, and I'd rather not delete the spring tests. Maybe we could merge this with a failing build? That'd still be better than no build at all, which is what we currently have.

composerinteralia:
Tests passed on Travis, and they pass locally on both MacOS and Linux with all the same versions of Ruby and gems. I have no idea what is going on here! 😢

composerinteralia:
I'm still totally stumped. In https://github.com/thoughtbot/factory_bot_rails/pull/396/commits/1e0de31f6053ffe2ddc86a7ed029cb20781d6b68 I can see that `Spring.project_root_path.join("config/application.rb").exist?` returns `true`, and so I wouldn't expect to see [this spring error](https://github.com/rails/spring/blob/v2.1.0/lib/spring/configuration.rb#L31). But once we run `bin/rails test` with spring enabled that's exactly the error we hit.

mathieujobin:
@composerinteralia it could be something like Github Actions does not set a current path
quite unusual to think of a process without `$PWD` but it looks like it ... 
maybe catch the error in place? fallback to a good default ?

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 3.0.2
```
mathieujobin, line 1: 3.0.4 ?

gemfiles/rails6.1.gemfile.lock
```diff
@@ -0,0 +1,267 @@
...
+      railties (>= 5.0.0)
+
+GEM
+  remote: https://rubygems.org/
+  specs:
+    actioncable (6.1.4.1)
+      actionpack (= 6.1.4.1)
+      activesupport (= 6.1.4.1)
+      nio4r (~> 2.0)
+      websocket-driver (>= 0.6.1)
```
mathieujobin, line 15: probably don't want to commit .lock in, and instead add `gemfiles/*lock` to the `.gitignore` file

.gitignore
```diff
@@ -1,6 +1,5 @@
 *.swp
 .bundle
-gemfiles/*.lock
```
mathieujobin, line 3: why?

composerinteralia:
I'm debugging in CI, since that's the only place it's failing. I'm not planning to merge any of these latest commits.

composerinteralia:
In my latest adventure, I really want to see this full error message: https://github.com/rails/spring/blob/v2.1.0/lib/spring/errors.rb#L25-L31, but I don't know how convince cucumber to show it to me. It's showing me the error class, not the full message.

mathieujobin:
I'll suggest we move the attempt to solve the spring issue to a later PR, disable the two tests. and hope someone will provide more insight on this issue later.

okuramasafumi:
Spring is not included as a default gem from Rails 7, so I think we can skip tests with Spring.
https://github.com/rails/rails/commit/21c9732fae9ce0fff7adf0da610d26eeedfc3076

mathieujobin:
yes, lets bring back this PR to something simple, and skip the few spring related issues.

gemfiles/rails6.1.gemfile.lock
mathieujobin, line 1: do not store .lock files on gems

composerinteralia: Some gems do, some don't. There are reasons for both choices. But in this case I only did it for debugging purposes.

composerinteralia:
Replaced by https://github.com/thoughtbot/factory_bot_rails/pull/410

PR #407, opened 2022-12-08
hack

mathieujobin:
Opening PR for visibility... we've been needing that for a year... I will try to remove see if can be trashed...

Thanks


composerinteralia:
This looks like a workaround for something specific to your application, and not something we'd merge into factory_bot_rails, so I'm going to close the PR.

PR #410, opened 2023-04-20
GitHub actions mathieu

mathieujobin:
replacement for #396 

just to get the ball rolling and have CI working again.
I don't use Spring, so I am skipping these two tests.

rubocop also complained, so I fixed the minimum to get a green build.

I didn't update the Appraisal file, honestly, I would remove it and instead have the Gemfile for more dynamic. I think this can be revisited by someone else in the future.

cheers


mathieujobin:
All tests passed 
https://github.com/mathieujobin/factory_bot_rails/actions/runs/4749285509/jobs/8436392670

composerinteralia:
Thanks for revisiting this. I had meant to come back and do this, but I forgot.

lib/factory_bot_rails/generators/null_generator.rb
```diff
@@ -1,9 +1,6 @@
 module FactoryBotRails
   module Generators
     class NullGenerator
-      def initialize(generators)
```
composerinteralia, line 4: This changes the arity of the initialize method, which I think means we'd fail on https://github.com/thoughtbot/factory_bot_rails/blob/57e88e79bf9d965eb4147cc111c2feff18b52c12/lib/factory_bot_rails/generator.rb#L12 if somebody has factory_bot_rails but adds configuration to disable the generators. I guess we don't have a test for disabling generators.

mathieujobin: ok, I can put back and add a rubocop exception. rubocop made me do it and I was dubious as well. rather not take the risk ...

.rubocop.yml
```diff
@@ -0,0 +1 @@
+inherit_from: .rubocop_todo.yml
```
composerinteralia, line 1: This project uses [standard](https://github.com/standardrb/standard), so I don't think we should have any rubocop config.

That might be a good thing to mention in the CONTRIBUTING.md (it's mentioned in the thoughtbot style guide that they use standard, but I don't think that's the first place somebody would look for that information).

mathieujobin: I'll remove them... I just noticed I had to do the ignores via standard...

mathieujobin:
All tests passed again ...

https://github.com/mathieujobin/factory_bot_rails/actions/runs/4751503955

.rubocop.yml
mathieujobin: I removed the rubocop files and added the ignore via standard-todo

composerinteralia (approved):
Many thanks!

PR #415, opened 2023-04-20
Cancel duplicate workflows

composerinteralia:
This avoids duplicate runs for both push and PR.

https://docs.github.com/en/actions/using-jobs/using-concurrency


composerinteralia:
Well that didn't work 🤷🏻‍♂️

PR #416, opened 2023-04-20
Expand matrix for more recent versions

composerinteralia:
This adds some extra coverage for combinations of newer versions of Ruby and Rails.

A number of the other versions are EOL or soon-to-be EOL, and we'll probably drop them anyway whenever we do another major release of factory_bot_rails, so being thorough there probably doesn't matter as much.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,11 +9,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rails: ["7.0"]
-        ruby: [3.2]
+        rails: ["7.0", "6.1"]
+        ruby: ["3.2", "3.1"]
         include:
-          - rails: "6.1"
-            ruby: "3.1"
           - rails: "6.1"
             ruby: "3.0"
           - rails: "6.1"


.github/workflows/build.yml
```diff
@@ -9,11 +9,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rails: ["7.0"]
-        ruby: [3.2]
+        rails: ["7.0", "6.1"]
+        ruby: ["3.2", "3.1"]
         include:
-          - rails: "6.1"
-            ruby: "3.1"
```
mathieujobin, line 16: this essentially changes nothing... rails 6.1, ruby 3.1 was already there

mathieujobin: Oh, I guess it adds 7.0/3.1 and 6.1/3.2

composerinteralia: Yeah, those are the ones I was going for. We'd probably have been fine without them, but it doesn't hurt.

mathieujobin: clearly, I like this pattern. I started using include instead of exclude recently. makes things much nicer. and with the quick matrix for the top two versions, its even better.

composerinteralia: I like it too!

PR #419, opened 2023-08-12
Reject sequence definitions for Active Record primary keys

seanpdoyle:
Alternative to [thoughtbot/factory_bot#1586][]
Depends on [thoughtbot/factory_bot#1587][]

First, introduce the `FactoryBotRails::FactoryValidator` class to serve as a generic `factory_bot.compile_factory` event observer.

Throughout the lifecycle of the `FactoryBotRails::Railtie`, afford various initializers with opportunities to add purpose-built validators to the instance's internal set.

When `factory_bot.compile_factory` events are published, iterate through the list of validators and forward along the value of the [event.payload][] to the validator's `#validate!` method.

Next, introduce the Active Record-specific
`FactoryBotRails::FactoryValidator::ActiveRecordValidator` class. Only require the module whenever [Active Record's engine is loaded][on_load].

The `ActiveRecordValidator#validate!` method rejects attributes that define primary key generation logic for `ActiveRecord::Base` descendants.

In order to test this behavior, add a development dependency on `sqlite3` and `activerecord`, along with some model and database table generating helper methods.

[thoughtbot/factory_bot#1586]: https://github.com/thoughtbot/factory_bot/pull/1586
[thoughtbot/factory_bot#1587]: https://github.com/thoughtbot/factory_bot/pull/1587
[event.payload]: module-ActiveSupport::Notifications-label-Subscribers
[on_load]: https://guides.rubyonrails.org/engines.html#avoid-loading-rails-frameworks

### DIFF
--- a/spec/support/macros/define_constant.rb
+++ b/spec/support/macros/define_constant.rb
@@ -1,0 +1,67 @@
+require "active_record"
+
+module DefineConstantMacros
+  def define_class(path, base = Object, &block)
+    const = stub_const(path, Class.new(base))
+    const.class_eval(&block) if block
+    const
+  end
+
+  def define_model(name, columns = {}, &block)
+    model = define_class(name, ActiveRecord::Base, &block)
+    create_table(model.table_name) do |table|
+      columns.each do |column_name, type|
+        table.column column_name, type
+      end
+    end
+    model
+  end
+
+  def create_table(table_name, &block)
+    connection = ActiveRecord::Base.connection
+
+    begin
+      connection.execute("DROP TABLE IF EXISTS #{table_name}")
+      connection.create_table(table_name, &block)
+      created_tables << table_name
+      connection
+    rescue Exception => e # rubocop:disable Lint/RescueException
+      connection.execute("DROP TABLE IF EXISTS #{table_name}")
+      raise e
+    end
+  end
+
+  def clear_generated_tables
+    created_tables.each do |table_name|
+      clear_generated_table(table_name)
+    end
+    created_tables.clear
+  end
+
+  def clear_generated_table(table_name)
+    ActiveRecord::Base
+      .connection
+      .execute("DROP TABLE IF EXISTS #{table_name}")
+  end
+
+  private
+
+  def created_tables
+    @created_tables ||= []
+  end
+end
+
+RSpec.configure do |config|
+  config.include DefineConstantMacros
+
+  config.before(:all) do
+    ActiveRecord::Base.establish_connection(
+      adapter: "sqlite3",
+      database: ":memory:"
+    )
+  end
+
+  config.after do
+    clear_generated_tables
+  end
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     factory_bot_rails (6.2.0)
-      factory_bot (~> 6.2.0)
+      factory_bot (~> 6.3.0)
       railties (>= 5.0.0)
 
 GEM
@@ -21,6 +21,11 @@ GEM
       erubi (~> 1.4)
       rails-dom-testing (~> 2.0)
       rails-html-sanitizer (~> 1.1, >= 1.2.0)
+    activemodel (7.0.4.3)
+      activesupport (= 7.0.4.3)
+    activerecord (7.0.4.3)
+      activemodel (= 7.0.4.3)
+      activesupport (= 7.0.4.3)
     activesupport (7.0.4.3)
       concurrent-ruby (~> 1.0, >= 1.0.2)
       i18n (>= 1.6, < 2)
@@ -69,7 +74,7 @@ GEM
     cucumber-tag-expressions (4.1.0)
     diff-lcs (1.5.0)
     erubi (1.12.0)
-    factory_bot (6.2.1)
+    factory_bot (6.3.0)
       activesupport (>= 5.0.0)
     ffi (1.15.5)
     ffi (1.15.5-java)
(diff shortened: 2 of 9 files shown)


Gemfile
```diff
@@ -8,3 +8,4 @@ gem "cucumber"
 gem "rake"
 gem "rspec-rails"
 gem "standard"
+gem "factory_bot", github: "thoughtbot/factory_bot", branch: "compile_factory_instrumentation"
```
seanpdoyle, line 11: Remove this line when https://github.com/thoughtbot/factory_bot/pull/1587 is merged and available as part of a release.

mike-burns (approved):
Nice, into it.

mike-burns:
The next steps, as I understand it:

1. merge the factory_bot PR
2. release factory_bot
3. remove the Gemfile modification here
4. merge this
5. release factory_bot_rails
6. announce it on our blog & fediverse

seanpdoyle:
The test suite passes locally, but fails in CI. I'm investigating.

Gemfile
seanpdoyle: > I assume this is because factory_bot isn't released yet.

@mike-burns I was hoping that this line would include the changes necessary to make CI pass. Do you know of a separate mechanism that's still loading the published gem?

mike-burns: Oh you're right, I overlooked that you're hardcoding the branch name in here.

I no longer have a guess as to why CI is failing!

mike-burns:
factory_bot 6.3.0 is out with support for the `compile_factory` notification.

seanpdoyle:
@mike-burns thanks for pushing that through! CI seems to be content with newer versions of Ruby, but failing on 2.7 and below. I'm not sure why.

spec/support/macros/define_constant.rb
```diff
@@ -0,0 +1,67 @@
+require "active_record"
+
+module DefineConstantMacros
```
mike-burns, line 3: Thanks for adding these.

mike-burns (approved):
Whew, looks great. Thanks for your patience while we figured out the right solution.

franzliedke:
@seanpdoyle This causes problems for us, as we have a table with a primary key that is not generated by the database (those two are not 100% the same thing).

Would it be possible to check the schema instead of the `primary_key` config? Something like this:

```ruby
ModelKlass.column_for_attribute(primary_key_name).serial?
```

This seems to work for me - not sure if this is Postgres-specific.

mike-burns:
Ah I see, we should check for `SERIAL`, not `PRIMARY KEY`. I think that makes sense -- do you agree, @seanpdoyle ?

seanpdoyle:
I agree, this conditional should account for whether or not the database is generating the value itself.

Unfortunately, I can't find any documentation for `#serial?`. There is a Postgres-specific [implementation](https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/connection_adapters/postgresql/column.rb#L15-L21) that's available in `activerecord>=5.2`, but it isn't part of the Column object's public interface.

Similarly, there is a [Column#auto_populated?](https://github.com/rails/rails/blob/7-1-stable/activerecord/lib/active_record/connection_adapters/column.rb#L67-L73) predicate defined as public interface that Postgres implements in terms of `#serial?`, but that's only available from `activerecord>=7.1`.

Is there another way to determine whether or not a column is managed by the database that works with older versions?

jagthedrummer:
> if you can think of a usecase then we'd like to know soon.

@mike-burns, I'm way late to the party, but I just wanted to chime in.

We have a usecase where we use FB to generate example JSON payloads (using `build` instead of `create`) and we want the example to contain IDs (and timestamps and other values that would be present in a "complete record").

For example, we have a factory defined like this:

```ruby
FactoryBot.define do
  factory :team do
    sequence(:name) { |n| "Generic Team #{n}" }
    sequence(:slug) { |n| "team_#{n}" }
    time_zone { nil }
    locale { nil }

    factory :team_example do
      id { 42 }
      sequence(:name) { |n| "EXAMPLE Generic Team #{n}" }
      sequence(:slug) { |n| "EXAMPLE team_#{n}" }
      time_zone { ActiveSupport::TimeZone.all.first.name }
      locale { "en" }
      created_at { DateTime.new(2023, 1, 1) }
      updated_at { DateTime.new(2023, 1, 2) }
    end
  end
end
```

And then to demonstrate what a `Team` JSON payload looks like we do something like this:

```ruby
team = FactoryBot.build(:team_example)
payload = team.as_json
```

Ideally we'd like for the `FactoryBot::AttributeDefinitionError` to be raised for the `:team` factory, but _not_ for the `:team_example` factory. So it would be nice to have a factory-by-factory way to preserve the previous behavior.

arcreative:
This one bit me too... just trying to figure out why this was released in a minor version when it's a breaking change for people who use sequences that also happen to be primary keys. 

In my particular use case, we use varchar primary keys that are imported from another system, but when running specs, we just coerce an integer to a string with an arbitrary prefix. For now, I've just used `factory_bot.reject_primary_key_attributes = false`, but would rather not, as we have almost a dozen projects where this would need to be implemented.

mike-burns:
I haven't forgotten about this. My plan is to revert the hard rejection and add it as a configurable lint. I'd also like it to only happen for serial/autoincrementing columns.

arcreative:
@mike-burns not sure how others feel about this, but I’m okay with the rejection as long as it doesn’t happen for primary keys that aren’t actually serial. As long as that’s reliable, I think the error is sensible enough.

mike-burns:
We should switch to using `ActiveRecord::ConnectionAdapters::Column#auto_populated?`. This method was [added to Rails](https://github.com/rails/rails/pull/48241) on June 1.

```
[1] pry(main)> User.column_for_attribute("email").auto_populated?
=> nil
[2] pry(main)> User.column_for_attribute("id").auto_populated?
=> "gen_random_uuid()"
```

As a second step, `sequence` should be able to opt in to overriding these auto-populated columns. Something like,

```ruby
sequence(:id, intentionally_override_auto_populated_database_column: true) { |n| "#{n}#{n}" }
```

mike-burns:
For now I've removed it: https://github.com/thoughtbot/factory_bot_rails/pull/451.

PR #429, opened 2023-10-27
CI against Rails 7.1

y-yagi:
This PR adds Rails 7.1 to the CI matrix to ensure the gem works with Rails 7.1.

This PR includes two commits to pass the test.

* [Ignore the directory that keeps factory files under the lib](https://github.com/thoughtbot/factory_bot_rails/pull/429/commits/2f6ab46973675161a143a0c008788ecf8e8b5c01)
    Some tests put factory files under the `lib`. But since Rails 7.1, Rails loads `lib` directory by default in a new application.
https://guides.rubyonrails.org/7_1_release_notes.html#introducing-config-autoload-lib-and-config-autoload-lib-once-for-enhanced-autoloading
  But factory files don't follow the naming rule of Zeitwerk. So Zeitwerk raises `Zeitwerk::NameError`. To avoid the error, this 
  changed to ignore the directory that puts factory files. 

* [Remove the route that is defined by default](https://github.com/thoughtbot/factory_bot_rails/pull/429/commits/1a48afd1f3bdbc7f70f404d5f3511e37fc658a65)
     Some test code loads the route file twice. This wasn't a problem before Rails 7.0 because we didn't have a route.
     But, Rails 7.1 has one route by default. So the test application raises the following error during the load.

    ```  
    Invalid route name, already in use: 'rails_health_check'  (ArgumentError)
    You may have defined two routes with the same name using the `:as` option, or you may be overriding a route already 
    defined by a resource with the same naming. For the latter, you can restrict the routes created with `resources` as explained 
    here:
    https://guides.rubyonrails.org/routing.html#restricting-the-routes-created
    ```

   The routing isn't a matter of this gem, so just ignore that for running the test.


mathieujobin (approved):

neilvcarvalho:
This PR's commits were cherry-picked to #440, and released on https://github.com/thoughtbot/factory_bot_rails/releases/tag/v6.4.2

Thank you for working on this!

Issue #433, opened 2023-11-17
undefined method `config' for nil:NilClass after upgrading to 6.4.0


julienchabanon:
### Description

After update to 6.4, when starting my rails app I get:
`/home/maraud/.rvm/gems/ruby-3.0.4/gems/railties-7.0.4/lib/rails.rb:47:in `configuration': undefined method `config' for nil:NilClass (NoMethodError)`

Downgrade to 6.2.X solve the issue

### Reproduction Steps

Upgrade factory bot rails to 6.4.0
Try to launch puma

### Expected behavior
The rails app should be able to initailize or steps on how to make it with 6.4 should be provided

### Actual behavior

`/home/maraud/.rvm/gems/ruby-3.0.4/gems/railties-7.0.4/lib/rails.rb:47:in `configuration': undefined method `config' for nil:NilClass (NoMethodError)`

### System configuration
**factory_bot_rails version**:  6.4.0
**factory_bot version**:  6.4.0
**rails version**:  7.0.4
**ruby version**: 3.0.4

julienchabanon:
https://github.com/thoughtbot/factory_bot_rails/pull/432 might solve the issue!

sealabcore:
Also experiencing this on 6.4.0 with Rails 7.1.2 and Ruby 3.1.4

tagliala:
Superseeds #426

rnestler:
Also experiencing this on 6.4.0 with Rails 7.0.8 and Ruby 3.2.2.

Workaround is to specify it with `gem 'factory_bot_rails', require: false` in the `Gemfile` and do a `require 'factory_bot_rails'` in `spec/rails_helper.rb`

kevinrobayna:
The work around just make it fail later on the factory internals 🤔 

here to be precise https://github.com/thoughtbot/factory_bot_rails/blob/3346305320df35eadae4ac942e44fa7adbbfa574/lib/factory_bot_rails/factory_validator/active_record_validator.rb#L7

tagliala:
> undefined method `<' for nil:NilClass

should have been tracked down and fixed in #431

rnestler:
> The work around just make it fail later on the factory internals 🤔

Make sure you updated `factory_bot` (not `factory_bot_rails`) to `6.4.2`.

akaspick:
[...]

PR #435, opened 2023-11-21
Change from `on_load :active_record` to `on_load :after_initialize` to ensure the Rails  configuration is available

TylerRick:
Otherwise one can run into this error:
```
  gems/railties-7.1.2/lib/rails.rb:51:in `configuration': undefined method `config' for nil:NilClass (NoMethodError)

      application.config
                 ^^^^^^^
  from gems/factory_bot_rails-6.4.0/lib/factory_bot_rails/railtie.rb:28:in `block in <class:Railtie>'
```


TylerRick:
Oops, I created this before I'd seen #433.

Looks like #432  (initializer) would be a better solution. Closing...

PR #439, opened 2023-11-23
Leoarnold/activerecord on load

adifsgaid:


adifsgaid:
Hello @mike-burns, I think this will cover the tests for the following [PR](https://github.com/thoughtbot/factory_bot_rails/pull/432), if anything needs to be changed, I will gladly do so 😊

Issue #438, opened 2023-11-22
Ability to disable reject_primary_key_attributes per factory


rmehner:
Newer versions of factory_bot/factory_bot_rails have this neat footgun preventing feature of rejecting sequences for primary key columns.

However, we have one use case where the primary key column is an externally provided string that is not an UUID, but just a string. With the newest update, our factories which look like this:

```rb
FactoryBot.define do
  factory :cms_item do
    name { "CMS item" }
    sequence(:id) { |n| n.to_s }
  end
end
```

will lead to tests failing with:

```
Attribute generates "id" primary key for CmsItem"
     
Do not define "id". Instead, rely on the database to generate it.
```

We can disable this with by setting `config.factory_bot.reject_primary_key_attributes` to `false`, however this will disable it for all factories. Since I'm not a fan of footguns, I'd love to have more fine grained control over this.

### Desired solution

I'd love to disable this per factory, kinda like this:

```rb
FactoryBot.define do
  factory :cms_item do
    name { "CMS item" }
	# this is an external id and we hope they know what they're doing
    do_not_reject_primary_key_attributes!
    sequence(:id) { |n| n.to_s }
  end
end
```

Naming obviously happy to be changed.

## Alternatives considered

I'm not hard stuck on the exact syntax being used, anything that allows me to have more fine grained control over this very nice feature is welcome. I don't want to get rid of it, since I have seen this gun hurting foots before 🚑 

## Additional context

Thank you for your work on this!

ybiquitous:
Hi. I think this validation should skip a case where the primary key isn't *auto-generated* by default, but what do you think?

https://github.com/thoughtbot/factory_bot_rails/blob/54c72541ac8d9afa2c94274d80597c24112ce054/lib/factory_bot_rails/factory_validator/active_record_validator.rb#L7

For example, [`ActiveRecord::Base.sequence_name`](https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-sequence_name) may be available for achieving it:

```diff
-if for_class < ActiveRecord::Base && for_class.primary_key == attribute.name.to_s
+if for_class < ActiveRecord::Base && for_class.primary_key == attribute.name.to_s && for_class.sequence_name
```

Although I'm unsure about an auto-increment primary key like MySQL.

tagliala:
Hello,

I'm here for the same issue. Assuming that all IDs are auto-generated is a strong assumption.

[...]

PR #444, opened 2023-12-14
Allow specific factories to skip primary key definition checks

zorab47:
Note: This PR is only a proof of concept idea due to monkey-patching `FactoryBot::Factory`.

## Problem

A new default in factory_bot_rails is to raise errors when a factory is defining a primary key. There are occasionally needs to assign a `ActiveRecord` model's primary key explicitly in a factory (see examples in #438). We don't want to entirely skip that additional check for every factory by disabling `reject_primary_key_attributes` for all factories.

## Possible Solution

Add a per factory configuration to skip the check for specific factories.

```ruby
FactoryBot.define do
  factory :warehouse, config: { allow_primary_key_definitions: true } do
    sequence(:id)
  end
end
```

### DIFF
--- a/lib/factory_bot_rails/factory_validator/active_record_validator.rb
+++ b/lib/factory_bot_rails/factory_validator/active_record_validator.rb
@@ -2,9 +2,17 @@ module FactoryBotRails
   class FactoryValidator
     class ActiveRecordValidator
       def validate!(payload)
-        attributes, for_class = payload.values_at(:attributes, :class)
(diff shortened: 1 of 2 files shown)


mike-burns:
Ah, I was thinking of passing through the config via the ActiveSupport::Notifications payload.

neilvcarvalho:
Closing this PR since it's a proof-of-concept for a changing a feature that is gone. Thank you for putting this out.

PR #451, opened 2023-12-30
Allow primary key attributes

mike-burns:
This improvement went out too quickly, so let's undo it.

Lots of the work around it was great and we want to keep that, so this surgically snips out:

- The README describing the feature.
- The ActiveRecordValidator that implements the validation.
- The railtie code that loaded the validator.

It leaves the validation framework and the notification code.


mike-burns:
I'd like to see this come back as a lint against [the `auto_populated?` predicate](https://github.com/thoughtbot/factory_bot_rails/pull/419#issuecomment-1847645604).

PR #473, opened 2024-05-24
Ignore Project IDX workspace customisations

smaboshe:
When using Google Project IDX, it sets up a local customisations folder (`.idx`). This commit ignores that folder.

This commit also sorts the Git ignore file to make it easier to scan.

Ref:
- https://idx.dev/
- https://developers.google.com/idx/guides/customize-idx-env

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,12 @@
 *.swp
 .bundle
-gemfiles/*.lock
+.idx
+.rubocop-https*
 coverage
+factory_bot_rails-*.gem
 factory_girl_rails-*.gem
+gemfiles/*.lock
 pkg
 rdoc
 test.db
 tmp
-.rubocop-https*
-factory_bot_rails-*.gem


sarahraqueld:
Have you considered adding the Google Project IDX files to your global .gitignore instead? This way, we don't need to include IDE-specific rules in each project's .gitignore file

smaboshe:
> Have you considered adding the Google Project IDX files to your [local .gitignore](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) instead? This way, you don't need to include IDE-specific rules in each project's .gitignore file:
> 
> > To always ignore a certain file or directory, add it to a file named ignore that's located inside the directory ~/.config/git.

Good call!

Yes, I have added this to my local (laptop) setup. The addition here is so that IDX ignores it as well. I'll see if there is a way to add a global `.gitignore` to IDX. I think that will be cleaner.

smaboshe:
I have managed to create a global `. gitignore` for the IDX environment.

PR #481, opened 2024-07-19
Add a dynamic security workflow and a SECURITY.md file

sarahraqueld:
We’re committed to protecting our systems, information, and our client’s information.

One of the ways we can enforce that in our OS projects is to provide a way to report a Vulnerability.

Add a dynamic security workflow that will automatically create PRs to update our Security policies in this repo.

### DIFF
--- a/.github/workflows/dynamic-security.yml
+++ b/.github/workflows/dynamic-security.yml
@@ -1,0 +1,19 @@
+name: update-security
+
+on:
+  push:
+    paths:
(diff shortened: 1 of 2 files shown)


rakeshpetit (approved):
We're only updating the workflow and md. So I hope the build errors are okay.

sarahraqueld:
> Details

Yes, I'm working on fixing the build errors in another PR.

PR #482, opened 2024-07-19
Fix standardrb violations

sarahraqueld:
Fixes https://github.com/thoughtbot/factory_bot_rails/issues/474

### Problem

Standard is failing: 

```standard: Use Ruby Standard Style (https://github.com/standardrb/standard)
spec/support/macros/define_constant.rb:10:40: Style/ArgumentsForwarding: Use anonymous block arguments forwarding (&).
spec/support/macros/define_constant.rb:11:52: Style/ArgumentsForwarding: Use anonymous block arguments forwarding (&).
spec/support/macros/define_constant.rb:20:32: Style/ArgumentsForwarding: Use anonymous block arguments forwarding (&).
spec/support/macros/define_constant.rb:25:43: Style/ArgumentsForwarding: Use anonymous block arguments forwarding (&).
```

Running `standardrb --fix` changes `&block` to `&`, however, this only works for ruby >= 3.2.

Since we still support Ruby 3.0, this change introduces these errors: 

```
SyntaxError:

  /home/runner/work/factory_bot_rails/factory_bot_rails/spec/support/macros/define_constant.rb:10: syntax error, unexpected ')', expecting local variable or method
  ...ne_model(name, columns = {}, &)
  ...                              ^
```

Issue #474, opened 2024-05-24
Fix standard errors


sarahraqueld:
On Github actions:

standard: Use Ruby Standard Style (https://github.com/standardrb/standard)
  spec/support/macros/define_constant.rb:10:40: Style/ArgumentsForwarding: Use anonymous block arguments forwarding (`&`).
  spec/support/macros/define_constant.rb:11:52: Style/ArgumentsForwarding: Use anonymous block arguments forwarding (`&`).
  spec/support/macros/define_constant.rb:20:32: Style/ArgumentsForwarding: Use anonymous block arguments forwarding (`&`).
  spec/support/macros/define_constant.rb:25:43: Style/ArgumentsForwarding: Use anonymous block arguments forwarding (`&`).
standard: Run `rake standard:fix` to fix up to 4 problems.

https://github.com/thoughtbot/factory_bot_rails/actions/runs/9221161554/job/25369617245?pr=468

smaboshe:
Nice catch!

Would it be worth adding a Standard check to our GitHub Actions?

PR #507, opened 2024-10-25
Drop support for Ruby 3.0

FerPerales:
According to [Ruby lang page](https://www.ruby-lang.org/en/downloads/branches/), Ruby 3.0 reached EOL on April 2024.

Supporting Ruby 3.0 is preventing the project from implementing standardrb most recent version, since we need to add an special case for syntax introduced in 3.2.

See #482 and #474

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         rails: ["7.1", "7.0", "6.1"]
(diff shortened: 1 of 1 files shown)


smaboshe (approved):
Thank you! This is a good reminder to watch the list of officially supported versions of Ruby.

sarahraqueld (approved):

sarahraqueld:
Thanks, Fer! https://github.com/thoughtbot/factory_bot_rails/issues/474 is now ready for review 🥳

PR #510, opened 2024-10-28
Add missing required Ruby version to gemspec

### LINKED ISSUE #456: gemspec is missing required_ruby_version
Closely related to https://github.com/thoughtbot/factory_bot/issues/1614: the gemspec doesn't claim any specific Ruby compatibility, but it's pretty clear from commits like https://github.com/thoughtbot/factory_bot_rails/commit/499635fcaffd5a1701626246716a74440970f23c that only specific ones are supported. The gem will be far easier to use if the gemspec can make that clear.

unused:
fixes #456

`required_ruby_version` is an optional property in specification though [recommended by Rubygems guide](https://guides.rubygems.org/specification-reference/). The change requests follows specification of [factory_bot](https://github.com/thoughtbot/factory_bot/blob/main/factory_bot.gemspec#L18).


sarahraqueld (approved):
Thank you, @unused !

PR #522, opened 2025-01-07
Add option to lazy load factory definitions

quadule:
Loading factory definitions can sometimes have a noticeable impact on Rails boot time: For example, in a large app I work on `FactoryBot.find_definitions` takes an average of 0.3s to load 274 factories. An option to lazy load them would avoid this overhead when factories are not needed.

The new `lazy_load_definitions` option is disabled by default for compatibility reasons, but all tests pass for both `factory_bot` and `factory_bot_rails` when enabled. The railtie will also continue to load all factory definitions at boot if `config.eager_load` is enabled in the application.

### DIFF
--- a/lib/factory_bot_rails/lazy_registry_find.rb
+++ b/lib/factory_bot_rails/lazy_registry_find.rb
@@ -1,0 +1,10 @@
+require "factory_bot/registry"
+
+module FactoryBotRails
+  module LazyRegistryFind
+    def find(*)
+      FactoryBot.find_definitions unless FactoryBot.factories.any?
+      super
+    end
+  end
+end
--- a/lib/factory_bot_rails/railtie.rb
+++ b/lib/factory_bot_rails/railtie.rb
@@ -10,6 +10,7 @@ module FactoryBotRails
   class Railtie < Rails::Railtie
     config.factory_bot = ActiveSupport::OrderedOptions.new
     config.factory_bot.definition_file_paths = FactoryBot.definition_file_paths
+    config.factory_bot.lazy_load_definitions = false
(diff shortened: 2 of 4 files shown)


lib/factory_bot_rails/railtie.rb
```diff
@@ -21,7 +22,13 @@ class Railtie < Rails::Railtie
     end
 
     config.after_initialize do |app|
-      FactoryBot.find_definitions
+      if app.config.factory_bot.lazy_load_definitions && !app.config.eager_load
+        require "factory_bot_rails/lazy_registry_find"
+        FactoryBot::Registry.prepend FactoryBotRails::LazyRegistryFind
```
quadule, line 27: This should probably be implemented in the `factory_bot` gem instead, so I'll open another PR for that if there's any interest here.

lib/factory_bot_rails/lazy_registry_find.rb
```diff
@@ -0,0 +1,16 @@
+require "factory_bot/registry"
+
+module FactoryBotRails
+  module LazyRegistryFind
+    def self.find_definitions_once
+      return if defined?(@definitions_loaded)
```
jrafanie, line 6: @quadule 👋 Fellow user of this gem... I like the concept of this PR... but, this concerns me, how does this interface with FactoryBot.reload? Does this caching get cleared on reload?

In other words, do we get a new registry when the configuration is reset in [reset_configuration](https://github.com/thoughtbot/factory_bot/blob/8446cb6c5b39ea046d8ba180197aabc66adf62ed/lib/factory_bot/reload.rb#L3)?

PR #526, opened 2025-01-31
Fix development dependencies

neilvcarvalho:
Some changes in our dependency tree have led to broken builds. For example, `concurrent-ruby`, which Rails uses, stopped requiring `logger` on version 1.3.5, which was fixed in Rails 7.1. Another change is that Ruby decreased its standard library, removing some preloaded gems, such as `mutex_m`.

This commit updates the dependencies to the latest supported version, limits the `concurrent-ruby` version to `< 1.3.4` in Rails versions that don't require `logger`, and adds `mutex_m` to the list of development dependencies.


DoodlingDev (approved):

spec/fake_app.rb
```diff
@@ -24,6 +24,6 @@ class Application < Rails::Application
   end
 end
 
-Rails.logger = Logger.new("/dev/null")
+Rails.logger = Logger.new(File::NULL)
```
DoodlingDev, line 27: nice, didn't know about this :scribbles in notes:

neilvcarvalho: TIL too - thanks new `standard` check

neilvcarvalho:
@seanpdoyle It will! That was happening because `concurent-ruby` stopped requiring `logger` (which `rails` didn't). It was only fixed in Rails 7.1 onwards.

PR #530, opened 2025-02-07
Run Cucumber feature tests on CI

neilvcarvalho:
The CI builds are just too fast. We're only running RSpec and Minitest there, skipping the much slower Cucumber feature tests. This has happened because we broke the RSpec and Minitest runs into two different CI steps, forgetting Cucumber was there.

This commit re-adds Cucumber as a separate step.


neilvcarvalho:
This will probably fail. And this explains why the tests are taking so long (and failing) locally, but passing so fast on CI.

PR #533, opened 2025-05-13
Only reload if cache_classes is disabled and reload_classes_only_on_c…

stanhu:
…hange disabled

https://guides.rubyonrails.org/configuring.html#config-reload-classes-only-on-change states:

>>>
3.2.30 config.reload_classes_only_on_change

Enables or disables reloading of classes only when tracked files change. By default tracks everything on autoload paths and is set to true. If config.cache_classes is true, this option is ignored.
>>>

It was surprising to us that even though `cache_classes` were set to true that factory_bot_rails still instantiated file watchers.


stanhu:
Hmm, this is not so easy to test directly because `require 'fake_app'` in `spec_helper.rb` always calls `after_initialize`, and that is only run once.

PR #538, opened 2025-06-17
Update `required_ruby_version` to require >= Ruby 3.1

y-yagi:
It seems that the support of Ruby 3.0 has dropped #507. This PR updates `required_ruby_version` to match it.


neilvcarvalho (approved):
Thank you!

PR #542, opened 2025-09-05
Add AuthenticationGenerator for users factory creation

neilvcarvalho:
I couldn't push to #540, so I'm opening a new PR with the original commit and a new one.

---

This change addresses an issue where running the `factory_bot` generator could result in a `factory_bot [not found]` error. The problem occurred due to missing generator definitions and template files.

With this update, the authentication generator is properly defined and includes the necessary templates, ensuring that the `users.rb` factory file is generated regardless of whether the project is using RSpec (`spec/factories`) or Minitest (`test/factories`). This guarantees consistent behavior and avoids generator errors in environments without a `spec` directory.

Previously, projects without a `spec` directory or with incomplete generator/template setup would fail when invoking `factory_bot` generators, making it impossible to scaffold the necessary factory files automatically.

- Added the missing `factory_bot:authentication` generator definition.
- Included `users.rb` template under `templates` directory.
- Implemented logic to detect the test framework and place the generated factory file in the correct directory (`spec/factories` or `test/factories`).
- Ensured generator runs successfully regardless of the presence of a `spec` directory.

### DIFF
--- a/lib/generators/factory_bot/authentication/templates/users.rb
+++ b/lib/generators/factory_bot/authentication/templates/users.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :user do
+    password { "password123" }
+    password_confirmation { "password123" }
+    email_address { "user#{SecureRandom.hex(3)}@example.com" }
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/generators/factory_bot/authentication/templates/users.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :user do
+    password { "password123" }
+    password_confirmation { "password123" }
+    email_address { "user#{SecureRandom.hex(3)}@example.com" }
+  end
+end
```
jeromedalbert, line 9: This file is never used, I made PR https://github.com/thoughtbot/factory_bot_rails/pull/544 to remove it.

